PR #4, opened 2019-12-03
Pull in additional task data from the task and unittest tables

ahal:
This adds the following metadata for tasks:
- kind
- tags
- groups (aka manifests)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mozci"
-version = "0.1.3"
+version = "0.1.4"
 description = ""
 authors = [
   "Andrew Halberstadt <ahal@mozilla.com>",


ahal:
Marco, I wanted to get your thoughts here.

Unfortunately this means we run 3 AD queries to fill in all the task information as each table has a unique key that we need (classification from treeherder, groups from unittest, and kind from task).

This PR implements it such that we run all 3 right off the bat, though maybe I should do something more clever. Either:
1. Only run each query as needed (seems a bit weird)
2. Run these task queries across a wider push range

Do you mind if I land this and we implement one or both of these in a follow-up?

mozci/queries/push_tasks_from_task.query
```diff
@@ -0,0 +1,13 @@
+from: task
+select:
+    - {name: id, value: task.id}
+    - {name: kind, value: task.kind}
+    - {name: tags, value: task.tags}
```
marco-c: What were you planning to do with the kind and tags info? I'm not too familiar with the fields, just saw an example now with https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/crJUWRrkQN-V7N1IJUV6Kg.

Nit: I see `kind` is part of `tags`, maybe we can just retrieve `tags`?

mozci/push.py
```diff
@@ -106,20 +106,37 @@ def tasks(self):
...
-            # Do a bit of data sanitization.
-            if any(a not in kwargs for a in ('label', 'duration', 'result', 'classification')):
-                continue
-
-            if kwargs['duration'] <= 0:
+        tasks = defaultdict(dict)
+
+        def add(data):
+            for task in data:
+                if 'id' not in task:
```
marco-c: Is this common with activedata or can we log a warning?

mozci/push.py
```diff
@@ -106,20 +106,37 @@ def tasks(self):
...
+        for task in tasks:
+            if not task.get('tags'):
                 continue
-
-            tasks.append(Task(**kwargs))
-
-        return tasks
+            task['tags'] = {t['name']: t['value'] for t in task['tags']}
+
+        # If we are missing one of these keys, discard the task.
```
marco-c: Same here, should we log a warning when we discard a task or is it too frequent?

PR #10, opened 2019-12-11
Store unittest result, test and manifest as part of test tasks

marco-c:
I was unable to test this normally, as in all pushes I tried to analyze, most tasks were skipped because of missing fields.

### DIFF
--- a/mozci/queries/push_tasks_from_unittest.query
+++ b/mozci/queries/push_tasks_from_unittest.query
@@ -3,6 +3,9 @@ select:
     - {name: id, value: task.id}
     - {name: kind, value: task.kind}
     - {name: _groups, value: run.suite.groups}
+    - {name: _result_group, value: result.group}
+    - {name: _result_test, value: result.test}
+    - {name: _result_ok, value: result.ok}
 where:
     and:
         - prefix: {repo.changeset.id: {$eval: rev}}
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -76,22 +76,38 @@ class TestResult:
 
 @dataclass
 class TestTask(Task):
-    results: List[TestResult] = field(default_factory=list)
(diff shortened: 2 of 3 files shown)


mozci/task.py
```diff
@@ -68,6 +68,9 @@ def get_artifact(self, path):
 
 @dataclass
 class TestTask(Task):
+    result_group: str = ""
+    result_test: str = ""
+    result_ok: bool = True
```
ahal: Do you think it would be worth creating a `Result` dataclass to store individual results? Or are just adding unnecessary overhead at that point.

mozci/queries/push_tasks_from_unittest.query
```diff
@@ -2,6 +2,9 @@ from: unittest
 select:
     - {name: id, value: task.id}
     - {name: groups, value: run.suite.groups}
+    - {name: result_group, value: result.group}
+    - {name: result_test, value: result.test}
+    - {name: result_ok, value: result.ok}
```
ahal: I'm a bit confused by what's happening here. Doesn't a single task have hundreds of results? Is this actually going to select the right thing?

marco-c: Right, we should turn it into a list when we build the task

mozci/task.py
marco-c: Done.

mozci/queries/push_tasks_from_unittest.query
marco-c: It is now a list of Results.

marco-c:
There is a problem with the current implementation:
- When loaded from AD, we have all test and manifest results, not only failures;
- When loaded from the errorsummary fall-back, we have all manifest results, but only test failures.

So this doesn't affect us for manifest-level, but it's something to keep in mind if we ever want to go down to test-level.

PR #12, opened 2019-12-12
Load groups from errorsummary artifact instead of ActiveData

ahal:


ahal:
Note this will continue to use ActiveData to fill in groups when possible. But if the key is missing, then it will grab groups from the errorsummary.log as a fallback.

PR #16, opened 2019-12-13
Make sure to only pay attention to the relevant 'action' when reading…

ahal:
… errorsummary.log

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -79,20 +79,24 @@ class TestResult:
 class TestTask(Task):
     """Subclass containing additional information only relevant to 'test' tasks."""
     _results: List[TestResult] = field(default=None)
+    _errors: List = field(default=None)
     _groups: List = field(default=None)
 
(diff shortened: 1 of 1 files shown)


mozci/task.py
```diff
@@ -69,48 +69,54 @@ def get_artifact(self, path):
...
 
 
 @dataclass
 class TestTask(Task):
+    """Subclass containing additional information only relevant to 'test' tasks."""
     _results: List[TestResult] = field(default_factory=list)
     _groups: List = field(default_factory=list)
 
-    def _load_errorsummary(self):
+    @memoized_property
```
marco-c: I didn't do this in my PR as I wanted to avoid using too much memory when you have thousands of tasks loaded

ahal: Good point.

We actually also `memoize` the `get_artifact` function. So no need to memoize this value again.. the cost of splitting it into lines isn't terrible.

PR #20, opened 2020-01-10
Add alternatives to the 'regressions' functions to get group-level regressions

### LINKED ISSUE #9: Build alternatives to the `regression` functions which work at the manifest level
Once #3 is fixed.
The logic (and so the code) should possibly be the same, just working on manifests rather than task.

marco-c:
Fixes #9


marco-c:
This was reviewed by @ahal in person at the All Hands.

PR #21, opened 2020-01-13
larger resultset

klahnakoski:
Changes for requesting large resultsets

Depends on https://github.com/mozilla/adr/pull/46

### DIFF
--- a/mozci/queries/push_tasks_from_unittest.query
+++ b/mozci/queries/push_tasks_from_unittest.query
@@ -2,6 +2,7 @@ from: unittest
 select:
     - {name: id, value: task.id}
     - {name: kind, value: task.kind}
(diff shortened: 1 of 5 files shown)


mozci/queries/push_tasks_from_unittest.query
```diff
@@ -2,7 +2,6 @@ from: unittest
 select:
     - {name: id, value: task.id}
     - {name: kind, value: task.kind}
-    - {name: _groups, value: run.suite.groups}
```
ahal: Was this intentional?

klahnakoski: @ahal  That was a mistake.  Although, that is big field (an array of all groups), which works out a a few gigabytes per query.

PR #30, opened 2020-02-03
Fix the push_revisions query (and remove the push_revision_count query)

marco-c:


mozci/queries/push_revisions.query
```diff
@@ -3,15 +3,14 @@ groupby:
     - {name: pushid, value: push.id}
 select:
     - {name: date, value: push.date, aggregate: min}
-    - {name: rev, value: changeset.id, aggregate: union}
+    - {name: revs, value: changeset.id, aggregate: union}
```
klahnakoski: maybe use `changeset.id12` for something shorter?

marco-c: Good idea, I'll file a follow-up issue for it as we need to make sure there is nothing relying on the full ID.

Issue #27, opened 2020-02-03
Use ActiveData 'table' format


marco-c:
It is more compact, should be faster and use less disk space in the ADR cache.

marco-c:
Note: we can only use for the "normal" queries, not the `destination: url` ones (https://github.com/mozilla/ActiveData/blob/dev/docs/jx_big_data.md#some-important-notes).

marco-c:
Partly done with #33, the rest of the queries will have to wait for https://github.com/mozilla/ActiveData/issues/186.

PR #33, opened 2020-02-03
Use 'table' format when possible

marco-c:
Ref #27

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -154,18 +154,18 @@ class Push:
             "_result_ok", "_result_group", "_result_test"
         )
 
-        def add(data):
-            if "header" in data:
-                data["data"] = [
+        def add(result):
+            if "header" in result:
+                result["data"] = [
                     {
                         field: entry[i]
-                        for i, field in enumerate(data["header"])
+                        for i, field in enumerate(result["header"])
                         if entry[i] is not None
                     }
-                    for entry in data["data"]
+                    for entry in result["data"]
                 ]
 
-            for task in data["data"]:
+            for task in result["data"]:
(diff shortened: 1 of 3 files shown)


mozci/push.py
```diff
@@ -155,7 +155,17 @@ def tasks(self):
...
+                data["data"] = [
+                    {
+                        field: entry[i]
+                        for i, field in enumerate(data["header"])
+                        if entry[i] is not None
+                    }
+                    for entry in data["data"]
+                ]
+
+            for task in data["data"]:
```
klahnakoski: rename `data` to `result`, if only of clarity

mozci/push.py
```diff
@@ -155,7 +155,17 @@ def tasks(self):
...
         def add(data):
-            for task in data:
+            if "header" in data:
+                data["data"] = [
+                    {
+                        field: entry[i]
+                        for i, field in enumerate(data["header"])
+                        if entry[i] is not None
+                    }
+                    for entry in data["data"]
```
klahnakoski: use `zip` instead if an index (`i`):

```
[
    {
        field: value
        for field, value in zip(data['header'], entry) 
        if value is not None
    } 
    for entry in data['data'] 
]
```

or, if you do not care about extra Nones:

```
[dict(zip(data['header'], entry)) for entry in data['data']]
```

mozci/push.py
marco-c: Done with b63faaf.

mozci/push.py
marco-c: Done with 018f2b3.

PR #39, opened 2020-02-07
When counting the distance between a push and a failure, stop at the first push with the failure

marco-c:
This is also adding two basic unit tests.

### DIFF
--- a/tests/test_push.py
+++ b/tests/test_push.py
@@ -1,0 +1,244 @@
+# -*- coding: utf-8 -*-
+
+from mozci.push import Push
+from mozci.task import Task
+
+
+def test_succeeded_in_parent_didnt_run_in_current_failed_in_child_failed_in_grandchild():
+    '''
+    Tests the scenario where a task succeeded in a parent push, didn't run in the
+    push of interest, and failed in its following pushes.
+    '''
+    first = Push("first")
+    parent1 = Push("parent1")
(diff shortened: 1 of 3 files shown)


tests/test_push.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
```
ahal: Not that I mind, but the license headers aren't technically needed per-file here. Also tests are generally just dedicated to the public domain anyway.

tests/test_push.py
```diff
@@ -0,0 +1,64 @@
...
+    last = Push("last")
+
+    current.child = child1
+    current.tasks = []
+
+    child1.child = child2
+    child1.tasks = [Task.create(id="1", label="test-prova", result="testfailed", classification="not classified")]
+
+    child2.child = last
+    child2.tasks = [Task.create(id="1", label="test-prova", result="success")]
```
ahal: Nice, stubbing out Push objects is much nicer than trying to mock ActiveData. Not sure why I thought it would be more complicated.

Eventually we might want to build some fixtures to make it easier to generate Push instances like this, but this is fine for now.

PR #47, opened 2020-02-11
Drop bad, non-actionable, groups

marco-c:


mozci/task.py
```diff
@@ -82,6 +100,29 @@ class TestTask(Task):
...
+        if self._groups is not None:
+            self._groups = [
+                group.replace("\\", "/")
+                for group in self._groups
+                if not is_bad_group(self.id, group)
+            ]
+
+        def update_group(result):
+            if result.group is not None:
+                result.group = result.group.replace("\\", "/")
```
ahal: Might be worth inlining [this file](https://searchfox.org/mozilla-central/source/python/mozbuild/mozpack/path.py#77) to `mozci.util` and using the `normpath` function.

mozci/task.py
```diff
@@ -82,6 +100,29 @@ class TestTask(Task):
...
+
+            return result
+
+        if self._results is not None:
+            self._results = [
+                update_group(result)
+                for result in self._results
+                if result.group is None or not is_bad_group(self.id, result.group)
```
ahal: Here we assert that `result.group is None`, then in `update_group` we make sure `result.group is not None`. I think I see why this is necessary, but might be cleaner if `is_bad_group` returned `True` if `result.group is None`. Then you could remove the check both here and in `update_group`. Though I guess you might not want to log that case.

mozci/task.py
```diff
@@ -19,6 +21,22 @@ class Status(Enum):
...
+
+
+def is_bad_group(task_id, group):
+    global bad_group_warned
+
+    bad_group = os.path.isabs(group) or group.startswith("file://") or group.startswith("Z:")
+
+    if not bad_group_warned and (bad_group or "\\" in group):
+        bad_group_warned = True
+        logger.warning(f"Bad group name in task {task_id}: {group}")
```
ahal: Rather than logging here, why not collect all "bad groups" and log the total number of them after the fact?
```
bad_tasks = {<task id>}
bad_groups = {<groups>}
...
self.logger.warning(f"{len(bad_tasks)} tasks contain bad groups:\n{\n.join(list(bad_groups)[:4]}\nand {len(bad_groups)-4} more")
```

mozci/task.py
marco-c: It looks like it doesn't change `\\` to `/`: `posixpath.normpath(normsep("part1\\part2")) == "part1\\part2"`

mozci/task.py
marco-c: The None checks might actually be removed after the changes from #50. I'll do that in a follow-up.

mozci/task.py
marco-c: The problem is, when do we log them?

ahal: At the end of `__post_init__` ?

ahal: I guess it could get called multiple times, so you'd have to clear the globals after each log..

ahal: Oh, nvm, I see what you're saying.. I guess it would have to be in the Push class. Yeah, maybe not worth it.

mozci/task.py
marco-c: Yep, done with a0c1505de9a578bbd689ceaf213d45883c529661.

mozci/task.py
ahal: Oh right.. that function only does modifications if `os.sep != '/'`, and since this is running on Linux it's leaving it alone. We could modify it I guess but I'm fine with just leaving the replaces in if you want.

marco-c: Hopefully we can remove this post-processing soon anyway, once the bug is fixed (and after a few weeks to have enough new data).

PR #50, opened 2020-02-11
Skip gathering groups for tasks which don't support groups yet

marco-c:
Depends on #47.

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -21,6 +21,25 @@ class Status(Enum):
     INTERMITTENT = 2
 
 
+NO_GROUPS_SUITES = (
+    "raptor",
+    "talos",
+    "awsy",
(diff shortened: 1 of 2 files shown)


mozci/task.py
```diff
@@ -100,9 +119,20 @@ class TestTask(Task):
     _errors: List = field(default=None)
     _groups: List = field(default=None)
 
-    # XXX: Once bug 1613937 and bug 1613939 are fixed, we can remove the filtering
-    # and slash replacing, and turn the warning on bad group names into an assertion.
     def __post_init__(self):
+        if is_no_groups_suite(self.label):
+            assert self._groups is None, f"{self.label} should have no groups"
```
ahal: Good idea!

Issue #38, opened 2020-02-07
Retrieve backedout information from ActiveData instead of HGMO


marco-c:
Depends on https://github.com/mozilla/ActiveData/issues/136.

marco-c:
This would help with https://github.com/mozilla/bugbug/issues/1449.

marco-c:
We are not using ActiveData anymore.

PR #52, opened 2020-02-12
Use 'fixed by commit' and back-out information to correctly attribute some regressions

### LINKED ISSUE #2: Use 'fixed by commit' and back-out information to correctly attribute some regressions
For example, test-macosx1014-64-shippable/opt-mochitest-devtools-chrome-e10s-1 is failing in https://treeherder.mozilla.org/#/jobs?repo=autoland&revision=ec8cad6891215a6db48cf2bc927ef6c77838035f.

The failure was marked by sheriffs as fixed by commit [6777cec97d18](https://hg.mozilla.org/integration/autoland/rev/69a232cb693503d98c04ddc5acace1d930bf4898), which is a backout of 31a1b40b20084e96947cf3b660a5b5ef9c5612d0.

test-macosx1014-64-shippable/opt-mochitest-devtools-chrome-e10s-1 did run on 31a1b40b20084e96947cf3b660a5b5ef9c5612d0, and it was successful.

marco-c:
Fixes #2

A nice example to test this on is ee582f48caa16183f496f63dfd85eefb288ba4fb, you can see the difference before/after.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -29,7 +29,7 @@ BASE_INDEX = "gecko.v2.{branch}.revision.{rev}"
 MAX_DEPTH = 14
 
 
-def get_hgmo(branch, rev):
+def get_automation_relevance(branch, rev):
     url = HGMO_AUTOMATION_RELEVANCE_URL.format(branch=branch, rev=rev)
     r = requests.get(url)
     r.raise_for_status()
@@ -46,7 +46,7 @@ def is_backout(branch, rev):
     if rev in backouts:
         return True
 
-    res = get_hgmo(branch, rev)
+    res = get_automation_relevance(branch, rev)
     if len(res["backsoutnodes"]) > 0:
         backouts.add(rev)
         return True
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -36,22 +36,17 @@ def get_automation_relevance(branch, rev):
     return r.json()["changesets"][0]
 
 
-backouts = set()
+backouts = {}
 
 
-@memoize
 def is_backout(branch, rev):
     global backouts
 
-    if rev in backouts:
-        return True
+    if rev not in backouts:
+        res = get_automation_relevance(branch, rev)
+        backouts[rev] = len(res["backsoutnodes"]) > 0
 
-    res = get_automation_relevance(branch, rev)
-    if len(res["backsoutnodes"]) > 0:
-        backouts.add(rev)
-        return True
-
-    return False
+    return backouts[rev]
 
 
 class Push:
@@ -88,7 +83,7 @@ class Push:
         if not backout_rev:
             return None
 
-        backouts.add(backout_rev)
+        backouts[backout_rev] = True
         return backout_rev
 
     @property
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -1,3 +1,4 @@
+import math
 from argparse import Namespace
 from collections import defaultdict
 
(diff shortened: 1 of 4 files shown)


mozci/push.py
```diff
@@ -27,6 +29,31 @@
...
+
+    if rev not in backouts:
+        return True
+
+    res = get_hgmo(branch, rev)
+    if len(res["backsoutnodes"]) > 0:
+        backouts.add(rev)
+        return True
+
+    return False
```
ahal: I think these should be methods on the `Push` class, and then `get_candidate_regressions` should instantiate a new `Push` instance and check `push.backsoutnodes`. Since everything is lazily loaded, there shouldn't be much additional perf impact.

You can set this up as a `memoized_property` the same way we do for `self._hgmo`.

mozci/push.py
```diff
@@ -27,6 +29,31 @@
 MAX_DEPTH = 14
 
 
+def get_hgmo(branch, rev):
```
ahal: The `Push` class already has a `self._hgmo` property. Let's call this `automation_relevance` instead.

mozci/push.py
```diff
@@ -17,6 +18,7 @@
 )
 from mozci.util.taskcluster import find_task_id
 
+HGMO_AUTOMATION_RELEVANCE_URL = "https://hg.mozilla.org/integration/{branch}/json-automationrelevance/{rev}"  # noqa
```
ahal: Unrelated to this patch, but I just noticed that the `Push` class won't work on `central` or `try` because we hardcode `integration` into the URL. I'll file an issue to fix this.

mozci/task.py
```diff
@@ -225,7 +226,7 @@ def __post_init__(self):
 
     @property
     def classifications(self):
-        return set(t.classification for t in self.tasks if t.failed)
+        return [(t.classification, t.classification_note) for t in self.tasks if t.failed]
```
ahal: Looks like we could move `def classifications` up to the `RunnableSummary` class.

mozci/push.py
marco-c: The only problem is that a push might contain multiple backouts, we'd have to load json-automationrelevance for all commits of a push in order to get all the backsoutnodes for a push (unless json-automationrelevance for the tip of a push gives us info about all commits of the push)

mozci/task.py
marco-c: Makes sense, I'll do that.

mozci/push.py
marco-c: #53.

mozci/push.py
marco-c: Here's an example push with three backouts: https://treeherder.mozilla.org/#/jobs?repo=autoland&revision=c078d08afa05139cdd28659604214a3bf79adde9.

I don't see a way out of this, the push revision here is c078d08afa05139cdd28659604214a3bf79adde9, but there are also 1a1332949d44be803d9c065399b0883cadd5ff8c and 63acd82c9360389a6dbd3bdf223e48bf32ca25c1.

If we have a failure where 'fixed by commit' is 1a1332949d44be803d9c065399b0883cadd5ff8c, we can't build a push instance with 1a1332949d44be803d9c065399b0883cadd5ff8c as there is no push with this revision. We'd have to first perform a request to get the push containing 1a1332949d44be803d9c065399b0883cadd5ff8c, then create a Push instance with the data returned by hgmo. So in the end we'd have too many requests to hgmo.

mozci/task.py
marco-c: Done with dac0242.

mozci/push.py
marco-c: Done with ba985d11558c5573ad938519273de61dfd5dc941.

mozci/push.py
ahal: Good point. Could we keep this out of `push.py` at least then? Maybe `util/hgmo.py` or something like that?

ahal: You no what, I have a patch in-progress that was moving the URL handling to a util file anyway, so just land this as is and I'll move it over later.

marco-c: Sounds good, and hopefully in the future we can just build a backout map right away using ActiveData (https://github.com/mozilla/mozci/issues/38).

PR #57, opened 2020-02-15
Use 'format: table' for the push revisions query

### LINKED ISSUE #27: Use ActiveData 'table' format

marco-c:
Fixes #27

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -570,12 +570,11 @@ class Push:
 
 
 def make_push_objects(**kwargs):
-    data = run_query("push_revisions", Namespace(**kwargs))["data"]
+    result = run_query("push_revisions", Namespace(**kwargs))
 
     pushes = []
 
-    for row in data:
-        pushid, date, revs, parents = row['pushid'], row['date'], row['revs'], row['parents']
+    for pushid, date, revs, parents in result["data"]:
         topmost = list(set(revs) - set(parents))[0]
 
         cur = Push([topmost] + [r for r in revs if r != topmost])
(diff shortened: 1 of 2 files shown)


mozci/push.py
```diff
@@ -570,12 +570,11 @@ def __repr__(self):
...
 
 def make_push_objects(**kwargs):
-    data = run_query("push_revisions", Namespace(**kwargs))["data"]
+    result = run_query("push_revisions", Namespace(**kwargs))
 
     pushes = []
 
-    for row in data:
-        pushid, date, revs, parents = row['pushid'], row['date'], row['revs'], row['parents']
+    for pushid, date, revs, parents in result["data"]["data"]:
```
marco-c: The `["data"]["data"]` is needed because of https://github.com/mozilla/ActiveData/issues/186#issuecomment-586573778 (normally it would just be one `["data"]`). If you want to change that on the ActiveData side, I'll update the PR.

klahnakoski: @marco, I will change it on the ActiveData side

klahnakoski: @marco-c this extra `data` appears to be happening inside ADR:  

https://github.com/mozilla/adr/blob/02cf57f2c05159f36b085910f427c2eff83ada33/adr/query.py#L149

The bulk `list` format coming from ActiveData is a simple array of objects, which ADR was so kind to turn into a standard query response.  The `table` format coming from ActiveData is a standard query response; ADR adds the extra `data` property.  
  
Since ADR must change for the table format, give me a little more time to "fix" the `list` format; so both bulk formats have a standard query response.

klahnakoski: I am waiting for test to pass in Activedata:  Meanwhile, here is an update to ADR

https://github.com/mozilla/adr/pull/59/files

marco-c: Thanks! I'll update the PR to match the new, correct, behavior.

Issue #53, opened 2020-02-13
hg.mozilla.org URLs are broken for release and try branches


ahal:
We hardcode `integration` into the HGMO urls:
https://github.com/mozilla/mozci/blob/c8648979e9b3dd7e0806e6d67df13e71eb20ce05/mozci/push.py#L20

This means we'll fail to download data there for central, try, release, etc. We should fix this :)

marco-c:
Reopening as #59 is being reverted for now (with #63).

PR #59, opened 2020-02-19
Refactor hg.mozilla.org url handling

ahal:
Creates a new HGMO class to handle all hg.mozilla.org activities.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -2,7 +2,6 @@ import math
 from argparse import Namespace
 from collections import defaultdict
 
-import requests
 from adr.errors import MissingDataError
 from adr.query import run_query
 from adr.util.memoize import memoize, memoized_property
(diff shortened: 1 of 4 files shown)


ahal:
Sorry for the force-pushing. I flagged you for review a bit prematurely.

ahal:
This fixes issue #53

mozci/util/hgmo.py
```diff
@@ -0,0 +1,66 @@
...
+    @staticmethod
+    def create(rev, branch='autoland'):
+        key = (branch, rev)
+        if key in HGMO.CACHE:
+            return HGMO.CACHE[key]
+        instance = HGMO(rev, branch)
+        HGMO.CACHE[key] = instance
+        return instance
+
+    @memoize
```
marco-c: I wonder if all the memoization is going to make us use too much memory. I guess we can worry about that if it ever happens.

mozci/push.py
```diff
@@ -78,12 +56,10 @@ def backedoutby(self):
         Returns:
             str or None: The commit revision which backs this push out (or None).
         """
-        global backouts
         backout_rev = self._hgmo.get('backedoutby')
         if not backout_rev:
             return None
 
-        backouts[backout_rev] = True
```
marco-c: This will make us perform more requests, not really so important though since we are going to switch to getting all backouts via ActiveData.

You might want to simplify this function since you're here, ```return self._hgmo.get('backedoutby') or None```

ahal: It actually won't because the `HGMO` class uses an instance cache. So it should be the same number, though as you noted previously, more memory consumption.

mozci/util/hgmo.py
ahal: I was thinking about this, and we should use `adr`'s caching mechanism (powered by [cachy](https://cachy.readthedocs.io/en/latest/)). That way we can store caches permanently on the file system, redis, memcached, etc.

Though the fs cache still isn't negligible, so there may be an argument to memoize it anyway. Maybe something like redis would be faster.

mozci/push.py
marco-c: In this case, we have a HGMO instance for the push revision, but not for the backout one and we were adding the backout revision to the "backouts" cache.
After the change, we'll have to get automation relevance for the backout revision.

ahal: Ah I see now, thanks! I can file an issue to add it back if you like, or if we'll be using AD soon then maybe not much point.

PR #60, opened 2020-02-20
Use a single session for HGMO requests

### LINKED ISSUE #37: Use fewer sessions so we don't have to re-connect often
We are creating a new session for each request. We should possibly re-use sessions so we don't have to reconnect often.
This is especially true for artifact downloads from S3.

marco-c:
Fixes #37


marco-c:
We'll need to reland #59 first before landing this.

PR #65, opened 2020-02-24
Implement some 'create_push' fixtures to reduce test boilerplate

ahal:


ahal:
@marco-c I only converted one test in the last commit as I think it's more of a matter of personal opinion whether that format is more readable or not.

So I'll either convert the rest of the tests or drop that commit entirely depending on what you prefer.

ahal:
I'll note that personally I like the new way better as it is more clear which push is which's parent at a glance.

marco-c:
I like the new way, much cleaner! The only drawback is that you need to think to figure out which one is the "push of interest" mentioned by the comment explaining the scenario.

ahal:
I could do something like:
```
p = create_pushes(6)
current = p[3]
```

I assume `current` was always the "push of interest"? We do make assertions on more than just that one, so technically aren't all pushes we make assertions against "of interest"?

marco-c:
Yeah I guess one can understand the test anyway, let's forget about it.

PR #73, opened 2020-03-02
Add some basic integration tests to make sure data we get from ActiveData is complete and matches our expectations

### LINKED ISSUE #71: Add some queries to make sure data from ActiveData is as expected
The way I see this working: we add the queries in mozci, we add a cron job somewhere (probably Taskcluster, or even Travis) to make sure the results from the queries are as we expect.

We could also do this in ci-recipes instead of mozci.

We could add one to make sure that all suites (except a blacklist) are generating manifest information. Something like:
```
{
	"from":"unittest",
	"groupby":"run.suite.name",
	"where":{"and":[
		{"eq":{"etl.id":0}},
		{"missing":"run.suite.groups"},
		{"gte":{"action.start_time":{"date":"today-1day"}}}
	]},
	"limit":1000
}
```

We could add another to make sure unittest results from all manifest-based suites (except a blacklist) have information on what manifest the result corresponds to. Something like:
```
{
	"from":"unittest",
	"groupby":"run.suite.name",
	"where":{"and":[
		{"exists":"run.suite.groups"},
		{"missing":"result.group"},
		{"gte":{"action.start_time":{"date":"today-1day"}}}
	]},
	"limit":1000
}
```

marco-c:
Fixes #71.


marco-c:
We need to decide whether we want to run this only on cron, or both on cron and on pushes/pull requests.

PR #78, opened 2020-03-09
Remove 'reftest' and 'crashtest' from NO_GROUPS_SUITE

ahal:


ahal:
I was hitting failures that these suites had groups. I think this is one of the ones :gbrown fixed. Just flagging you as I'm wondering why you hadn't hit it yet?

marco-c:
I've done the same change a few days ago: https://github.com/mozilla/mozci/commit/84c0af62b04807b6463e6515684ab37108758178.

Issue #51, opened 2020-02-11
Create documentation and README


ahal:
We should update the README and add some Sphinx documentation.

ahal:
Fixed by:
https://github.com/mozilla/mozci/commit/aef66e66fac686aaa6f5b1a2d0b2834b1e79f247

(Still tons of room for improvement, but at least we have a place we can start adding to)

PR #84, opened 2020-03-12
Fixes for ETL

klahnakoski:
Some points where None/missing data is found.  These may not be good solutions, but at least they point out where some problems are

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -282,8 +282,11 @@ class Push:
         Returns:
             set: A set of task labels (str).
         """
-        tasks = self.decision_task.get_artifact("public/task-graph.json").values()
-        return {t["label"] for t in tasks}
+        try:
+            tasks = self.decision_task.get_artifact("public/task-graph.json").values()
+            return {t["label"] for t in tasks}
+        except Exception as e:
+            return set()
 
     @property
     def unscheduled_task_labels(self):
@@ -534,6 +537,8 @@ class Push:
         """
         index = self.index + ".source.shadow-scheduler-{}".format(name)
         task = Task(id=find_task_id(index))
+        if task.id==None:
+            return set()
         labels = task.get_artifact("public/shadow-scheduler/optimized_tasks.list")
         return set(labels.splitlines())
 
--- a/mozci/util/taskcluster.py
+++ b/mozci/util/taskcluster.py
@@ -4,6 +4,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import functools
+import logging
 
 import requests
 import taskcluster_urls as liburls
@@ -86,6 +87,7 @@ def find_task_id(index_path, use_proxy=False):
         response = _do_request(get_index_url(index_path))
     except requests.exceptions.HTTPError as e:
         if e.response.status_code == 404:
-            raise KeyError("index path {} not found".format(index_path))
+            logging.getLogger(__name__).warning("index path {} not found".format(index_path))
+            return None
         raise
     return response.json()["taskId"]


mozci/util/taskcluster.py
```diff
@@ -86,6 +87,7 @@ def find_task_id(index_path, use_proxy=False):
         response = _do_request(get_index_url(index_path))
     except requests.exceptions.HTTPError as e:
         if e.response.status_code == 404:
-            raise KeyError("index path {} not found".format(index_path))
+            logging.getLogger(__name__).warning("index path {} not found".format(index_path))
+            return None
```
ahal: I think it's better to raise here and let consumers handle the exception. Maybe that's in the `push.py` file, or maybe it's in the `scheduler_analysis` script. I'd need to think about it.

mozci/push.py
```diff
@@ -282,8 +282,11 @@ def scheduled_task_labels(self):
         Returns:
             set: A set of task labels (str).
         """
-        tasks = self.decision_task.get_artifact("public/task-graph.json").values()
-        return {t["label"] for t in tasks}
+        try:
+            tasks = self.decision_task.get_artifact("public/task-graph.json").values()
+            return {t["label"] for t in tasks}
+        except Exception as e:
+            return set()
```
ahal: This seems dangerous. We want this to fail if there is an Exception, otherwise the errors might be going undetected.

mozci/push.py
```diff
@@ -534,6 +537,8 @@ def get_shadow_scheduler_tasks(self, name):
         """
         index = self.index + ".source.shadow-scheduler-{}".format(name)
         task = Task(id=find_task_id(index))
+        if task.id==None:
+            return set()
```
ahal: Do you remember the exception being raised here?

ahal:
I filed https://github.com/mozilla/mozci/issues/85

I'll leave this open for now just to remember where the pain points were.

mozci/push.py
klahnakoski: This line is propagating the Nones from the `if e.response.status_code == 404:` from `find_task_id()`

mozci/push.py
klahnakoski: This is a philosophical difference:  You are assuming you can contain the multitude of problems this code will have in production, and you can handle them all.  I do not care: At this point the artifact does not exist or is corrupted, or the programming is bad; I must proceed with a decision: There are no tasks scheduled for this push.

mozci/util/taskcluster.py
klahnakoski: That is up to you.  I do not intend for this PR to be the real solution.  Just a pointer to what needs to be handled.  How you handle it is up to you.

That said `find_task_id()` should return `None` when the task id does not exists.  Which is the case here.

mozci/util/taskcluster.py
```diff
@@ -4,6 +4,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import functools
+import logging
```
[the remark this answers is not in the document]
klahnakoski: I do not even want to emit logging, I was only using it to see the cause.

klahnakoski:
@ahal 

> In general, if `mozci` can't do something for some reason I consider it a feature that we fail loudly and raise an exception. It should be up to consumers to decide which exceptions they want to catch + ignore or which ones they want to leave alone.

I will leave the proper handling of exceptional cases to you.  In the case of ETL:  Functions can not be throwing exceptions because they are are part of a larger processes, or a larger expression, which has even less context about what is happening.  If a value is requested, and it makes no sense to ask for that value (eg asking for scheduled tasks on a decision task that does not exist) then return `None` .

> That said this PR is still useful, because we should be explicit about which exceptions could potentially be raised. Mozci should document what it expects (see issue #51). It should likely also define its own purpose built exception classes.

For my use cases, all those exception classes will collapse to `None`
 
> I'll see if I can get some documentation going, including API docs that include which exceptions can be raised from which functions/attributes.

Please put the documentation near the method headers. I would prefer to see docs where I am rather than somewhere else.   You seem to be doing this already, so I am happy.

ahal:
Understood that from an ETL's perspective, you just need to make sure that things keep chugging along. Though there will be many consumers of `mozci`. E.g, when Marco and I are running scripts to test things out, we need to fail loudly to detect when things aren't going as expected. If we use blanket exceptions and log a warning, it'll become very difficult to debug.

So I think the ETL will need to catch exceptions and handle them appropriately. I still think there is value in trying to distinguish the different types of "exceptional events". E.g, the ETL may decide to discard an entire push on certain exceptions, or it may decide to process the push anyway but with a default value.

ahal:
I was hoping to get started on this today, but the day has disappeared somehow. So I'd recommend using a blanket `except: Exception` wherever needed for now, and I'll try and get proper exceptions raised on Monday.

mozci/push.py
ahal: So I was in the process of creating a `MissingArtifactError`, but then realized that the `HTTPError` that requests raises is probably more informative. So why bother catching it and then immediately re-raising? I'll add the fact that `requests.HTTPError` can be raised to the appropriate docstrings and probably call this particular instance a day. I'll see if custom exceptions make more sense in other places.

klahnakoski: This is not important:

So, `scheduled_task_labels` is an attribute that can throw an `HTTPError`?  This is a low level error given the caller just wants a list of labels.  I do not think the caller should be concerned with **how** this attribute manages to screw up getting a list of labels.  In the same theme, we should not ask the caller to handle http connection setup errors, http download timeouts, or any number of http codes; at some point we bundle many reasons for failure into a single, actionable exception:  `MissingArtifactError` is a good choice.  The caller does not care **why** the artifact can not be had, it only needs to know there is no artifact.

Now, maybe you can set the cause: `MissingArtifactError().__cause__ = HTTPError`; then a smarter caller can still make sophisticated decisions.  Dumb callers can simply mitigate the lack of artifact. 

In general, from a data perspective, there is data or there is not:  In many cases `None` is an appropriate response. In all other cases an exception can be raised; an exception that only describes what the caller was requesting.

```
def get_tasks():
    try:
        # GET TASKS FROM SERVICE, OR CACHE
    except Exception as e:
        raise CanNotGetTasksException(
            "can not get tasks for push {{push}}", 
            push=self.id, 
            cause=e
        )
```

With this pattern, callers can switch behaviour on whether tasks can be had or not (which is most likely), and even switch on causes if they really want (so we never loose the ability to perform fine-grained decision making).

Using this pattern, I have never encountered a method that needs to throw a multiple exception types; multiple reasons are always bundled into one exception indicating failure of the method.

ahal: Yeah, ok you've convinced me again. That's what I was initially thinking, then started second guessing myself. It's possible to reraise an exception whilst keeping the original stacktrace which seems like what we'd want here.

klahnakoski: I believe python3 has better causal support.  Long ago I built libraries to keep stacktraces and causes for python2, and I have never looked into what Python3 is capable of now.  Sorry I am not much actual help.

PR #89, opened 2020-03-16
Add some basic Sphinx documentation

ahal:


ahal:
The CI failure is because `poetry` doesn't recognize `adr == 0.17.6` yet (I bumped it recently. It should sort itself out in time.

PR #91, opened 2020-03-17
Stop running the integration tests as part of the default test set

### LINKED ISSUE #90: Don't run integration tests per-commit
Marco added some integration tests, but we should likely not run them per commit as they can fail for reasons completely unrelated to the patch.

Instead, I think a good approach would be to run the tests in a TravisCI cron task that runs daily. We set something up similar in `adr`:
https://github.com/mozilla/adr/blob/master/tox.ini#L20

It also requires some configuration on the TravisCI side of things.

ahal:
Fixes #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,3 +51,10 @@ $ tox
 
 This should run successfully prior to submitting PRs (unless you need help figuring out the
 problem).
+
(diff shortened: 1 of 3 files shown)


ahal:
I also updated the cron task to run daily no matter what (even if there were builds).

PR #94, opened 2020-03-23
Handle `Push.rev` and `Push.revs` consistently

ahal:
The motivation here is to enforce more guarantees on the `rev` and `revs` properties. That it is the long form in the former, and that it exists at all in the latter.

This could potentially slow down performance, as it means we query `json-automationrelevance` in the `Push` constructor. I plan on using this fact to detect non-existent pushes early in a follow-up PR.

I theorize that we pretty much always need to query `json-automationrelevance` anyway, so my hope is that the overall perf impact is minimal.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -33,19 +33,30 @@ class Push:
 
     def __init__(self, revs, branch="autoland"):
         if isinstance(revs, str):
+            self._revs = None
             revs = [revs]
+        else:
+            self._revs = revs
 
-        self.revs = revs
         self.branch = branch
-        self._hgmo = HGMO.create(self.rev, branch=self.branch)
-        self.index = BASE_INDEX.format(branch=self.branch, rev=self.rev)
+        self._hgmo = HGMO.create(revs[0], branch=self.branch)
 
         self._id = None
         self._date = None
 
+        # Need to use full hash in the index.
+        if len(revs[0]) == 40:
+            self.rev = revs[0]
+        else:
+            self.rev = self._hgmo["node"]
+
+        self.index = BASE_INDEX.format(branch=self.branch, rev=self.rev)
+
     @property
-    def rev(self):
-        return self.revs[0]
+    def revs(self):
+        if not self._revs:
+            self._revs = [c["node"] for c in self._hgmo.changesets]
+        return self._revs
 
     @memoized_property
     def backedoutby(self):
--- a/mozci/util/hgmo.py
+++ b/mozci/util/hgmo.py
@@ -18,14 +18,9 @@ class HGMO:
     CACHE = {}
 
     def __init__(self, rev, branch="autoland"):
-        self.rev = rev
-        self.branch = branch
-
         self.context = {
-            "branch": "integration/autoland"
-            if self.branch == "autoland"
-            else self.branch,
-            "rev": self.rev,
+            "branch": "integration/autoland" if branch == "autoland" else branch,
+            "rev": rev,
         }
 
     @staticmethod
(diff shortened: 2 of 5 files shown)


marco-c:
> I theorize that we pretty much always need to query json-automationrelevance anyway, so my hope is that the overall perf impact is minimal.

Right now we are querying it only for getting backout information of some pushes if I'm not mistaken. After #38, we wouldn't use it at all.

What if we didn't allow building a Push with a single rev but always with a list of revs? In all usages in mozci we are always passing a list of revs (except in tests).

ahal:
The problem with that is it makes it very difficult to instantiate a `Push` object manually (I do this all the time in e.g ipython).

I guess I could move `revs` back behind an `@property` and only load it if `revs` *wasn't* passed in via the constructor. Though this still creates the opportunity for users/automation to pass in an invalid set of revs.

I'd still like to solve the problem of the shortened 'rev' though.. Would you be opposed to querying the regular `hgmo` data in the constructor? We use that for more things, including `push.parent` / `push.child`. So it seems much more likely we'd need that anyway.

marco-c:
> The problem with that is it makes it very difficult to instantiate a Push object manually (I do this all the time in e.g ipython).

I guess I could move revs back behind an @property and only load it if revs wasn't passed in via the constructor. Though this still creates the opportunity for users/automation to pass in an invalid set of revs.

What if we added a static function to create a push object given a single rev?
Or, if we want to make single rev the default, offer a way to avoid the HGMO load when we already have the list of revs (maybe revs could be an optional parameter of the constructor).

> I'd still like to solve the problem of the shortened 'rev' though.. Would you be opposed to querying the regular hgmo data in the constructor? We use that for more things, including push.parent / push.child. So it seems much more likely we'd need that anyway.

When we get the pushes from ActiveData, we cache the parents and children: https://github.com/mozilla/mozci/blob/66db80c5aba69d1aca6cda3231c8933b845d6fd6/mozci/push.py#L563-L568. So we only ever actually call parent for the very first pushes returned by ActiveData and child for the very last pushes.
So probably better to use automationrelevance so that we at least also cache the is_backout info.

ahal:
> Or, if we want to make single rev the default, offer a way to avoid the HGMO load when we already have the list of revs (maybe revs could be an optional parameter of the constructor).

This is what the latest revision of this PR is doing already :)

> So we only ever actually call parent for the very first pushes returned by ActiveData and child for the very last pushes.
> So probably better to use automationrelevance so that we at least also cache the is_backout info.

Ok, I guess we could pretty easily only do this if the revision is less than the full 40 characters. So it wouldn't have any perf impact as long as the full hash is used. It's important to do this because otherwise, the `self.index` is invalid and we get weird errors when trying to download artifacts and things.

mozci/push.py
```diff
@@ -33,19 +33,26 @@ class Push:
...
-        self.revs = revs
+        rev = revs[0]
         self.branch = branch
-        self._hgmo = HGMO.create(self.rev, branch=self.branch)
-        self.index = BASE_INDEX.format(branch=self.branch, rev=self.rev)
+        self._hgmo = HGMO.create(rev, branch=self.branch)
 
         self._id = None
         self._date = None
+        self.rev = self._hgmo['node']
```
marco-c: > This is what the latest revision of this PR is doing already :)

Isn't this going to trigger a load from hgmo?

ahal: Yes, I meant it does the `revs` thing on demand already.. The latest push is the one that also only gets `self.rev` from hgmo optionally.

ahal:
I was kind of hoping this PR would give us a way to verify that the `Push` is valid directly in the constructor. But since we don't hit `hgmo` in all cases, we won't be able to tell until later. I guess that will have to do.

Issue #85, opened 2020-03-12
Ensure we raise well-documented exceptions in each function


ahal:
A lot of things can go wrong in the process of downloading artifacts/querying AD/etc. We should make sure that each API surface only raises well-defined exceptions that are listed in our API docs (see issue #51).

This allows consumers to decide which errors to catch + ignore, and which ones to leave alone.

ahal:
This is more of an ongoing thing we should strive to do at this point, so closing the issue.

PR #95, opened 2020-03-23
Raise custom exceptions for common failure cases

ahal:
Currently we don't do any exception handling, resulting in a mixture of confusing `KeyError`s and `requests.HTTPError`s being raised to the user.

We should catch these and re-raise them with Python 3's `raise from` syntax:
```
try:
   self._hgmo.get('rev')
except requests.HTTPError as e:
    if e.response.status_code == 404:
        raise PushNotFoundException from e
    raise
```

### DIFF
--- a/mozci/errors.py
+++ b/mozci/errors.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+
+
+class BasePushException(Exception):
+    def __init__(self, rev, branch, msg):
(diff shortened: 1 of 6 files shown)


ahal:
The first two commits were supposed to be part of the other PR. I guess I did something wrong.. Was trying to make this PR based on the other one.

ahal:
This partially fixes issue #85.

PR #96, opened 2020-03-24
Switch from memoize to lru_cache

armenzg:
How do we know that it is working as expected?

On another note, look at [the second commit](https://github.com/mozilla/mozci/commit/5b1c326bbaa4c8793f4f5d6b3bcc32c2532bd045) and the output below.
I moved `@lru_cache` to the lower level function since it did not like using lru_cache as part of a function of a class.

```python
    def test_missing_artifacts(responses, create_task):
        artifact = "public/artifact.txt"
        task = create_task(label="foobar")
    
        # First we'll check the new deployment.
        responses.add(
            responses.GET, get_artifact_url(task.id, artifact), status=404,
        )
    
        # Then we'll check the old deployment.
        responses.add(
            responses.GET,
            get_artifact_url(task.id, artifact, old_deployment=True),
            status=404,
        )
    
        with pytest.raises(ArtifactNotFound):
>           task.get_artifact(artifact)
E           TypeError: unhashable type: 'Task'

tests/test_task.py:40: TypeError
============================================================================= short test summary info =============================================================================
FAILED tests/test_task.py::test_missing_artifacts - TypeError: unhashable type: 'Task'
```

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -3,10 +3,11 @@ import concurrent.futures
 import math
 from argparse import Namespace
 from collections import defaultdict
+from functools import lru_cache
 
 from adr.errors import MissingDataError
 from adr.query import run_query
-from adr.util.memoize import memoize, memoized_property
+from adr.util.memoize import memoized_property
 from loguru import logger
 
 from mozci.task import GroupResult, GroupSummary, LabelSummary, Status, Task, TestTask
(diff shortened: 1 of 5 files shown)


armenzg:
If you're looking for the Travis run showing up in this PR (as I was) here's the run:
https://travis-ci.org/github/mozilla/mozci/builds/666407038

Could it be that because I'm not a member I can't see it?

ahal:
I noticed Travis doesn't seem to show up sometimes.. I think it happens if you force-push in the middle of a build or something like that... But I don't see it either.

PR #109, opened 2020-03-26
[WIP] Penalize regressions that still occur on or after the backout of a push

### LINKED ISSUE #104: Penalize regressions that still occur on the backout of a push
If a failure is occurring on a push and on its backout, it was probably not caused by the push itself.

Note we can't totally exclude it, as there could be another push in between that causes another failure in the same label/group.

marco-c:
Fixes #104

A few observations:
- I'm not sure exactly what kind of penalization we should use. Similarly with the other penalizations, they have mostly been chosen without measurement. Possibly we should find a way to measure their effectiveness. Maybe we should build a "ground truth" dataset on which we can measure improvements.
- This works better at the group-level, since it's less likely that two pushes which are close together both introduced a regression in the same group. At the task-level, instead, the probability is higher. Maybe even when getting regressions at the task-level, we should consider the underlying groups to check if two failures are the same. Or, we can just ignore it since at some point we'll be group-level completely.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -491,17 +491,36 @@ class Push:
 
             total_count = count + child_count
 
-            # When the push was not backed-out, it's less likely to be the cause of a failure.
-            # So, we penalize it by doubling its count (basically, we consider the push to be
-            # further away from the failure, which makes it more likely to fall outside of
-            # MAX_DEPTH).
-            # We can't fully exclude pushes which were not backed-out because of bustage fixes.
-            if not self.backedout:
-                total_count *= 2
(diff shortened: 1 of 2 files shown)


ahal:
Did you still want me to review this @marco-c ? I need to figure out how to get better review notifications.

marco-c:
I'll remove it from your queue, I want to do this, but I need some time to evaluate it and it's not top priority.

marco-c:
I'm closing this as I'm implementing it as part of another PR (with a few fixes related to #198).
I'm not so concerned anymore about the problem of two close pushes causing a regression in the same task/group, as it should be pretty rare at the group-level and that's what we're focusing on.

PR #112, opened 2020-03-26
Fixes #86 - Cache task groups and results

### LINKED ISSUE #86: Cache task groups and results
When data is missing from ActiveData, we retrieve it from Taskcluster. We memoize it, but don't cache it, which means we have to get it again if we re-run.

armenzg:
I run the test locally with `pytest tests/test_integration.py -k test_caching_of_tasks`.
A directory `adr_cache` is created.

How can I determine if anything is being cached?

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -229,7 +229,8 @@ class Push:
         index = self.index + ".taskgraph.decision"
         return Task.create(index=index)
 
-    @memoized_property
+    # TODO: I removed memoized_property because the tests were failing with:
+    # TypeError: 'list' object is not callable
     def tasks(self):
         """All tasks that ran on the push, including retriggers and backfills.
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -229,8 +229,7 @@ class Push:
         index = self.index + ".taskgraph.decision"
         return Task.create(index=index)
 
-    # TODO: I removed memoized_property because the tests were failing with:
-    # TypeError: 'list' object is not callable
+    @memoized_property
     def tasks(self):
         """All tasks that ran on the push, including retriggers and backfills.
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -22,6 +22,122 @@ when the task did not run on the currently considered push.
 """
 
 
+@lru_cache(maxsize=None)
+def tasks(rev, branch):
+    args = Namespace(rev=rev, branch=branch)
+    tasks = defaultdict(dict)
+    retries = defaultdict(int)
+
+    list_keys = (
+        "_result_ok",
+        "_result_group",
+    )
+
+    def add(result):
+        if "header" in result:
+            result["data"] = [
+                {
+                    field: value
+                    for field, value in zip(result["header"], entry)
+                    if value is not None
+                }
+                for entry in result["data"]
+            ]
+
+        for task in result["data"]:
+            if "id" not in task:
+                logger.trace(f"Skipping {task} because of missing id.")
+                continue
+
+            task_id = task["id"]
+
+            # If a task is re-run, use the data from the last run.
+            if "retry_id" in task:
+                if task["retry_id"] < retries[task_id]:
+                    logger.trace(f"Skipping {task} because there is a newer run of it.")
+                    continue
+
+                retries[task_id] = task["retry_id"]
+
+                # We don't need to store the retry ID.
+                del task["retry_id"]
+
+            cur_task = tasks[task_id]
+
+            for key, val in task.items():
+                if key in list_keys:
+                    if key not in cur_task:
+                        cur_task[key] = []
+
+                    cur_task[key].append(val)
+                else:
+                    cur_task[key] = val
+
+    # Gather information from the treeherder table.
+    try:
+        add(run_query("push_tasks_from_treeherder", args))
+    except MissingDataError:
+        pass
+
+    # Gather information from the unittest table. We allow missing data for this table because
+    # ActiveData only holds very recent data in it, but we have fallbacks on Taskcluster
+    # artifacts.
+    # TODO: We have fallbacks for groups and results, but not for kind.
+    try:
+        add(run_query("push_tasks_results_from_unittest", args))
+    except MissingDataError:
+        pass
+
+    try:
+        add(run_query("push_tasks_groups_from_unittest", args))
+    except MissingDataError:
+        pass
+
+    # If we are missing one of these keys, discard the task.
+    required_keys = (
+        "id",
+        "label",
+    )
+
+    # Normalize and validate.
+    normalized_tasks = []
+    for task in tasks.values():
+        missing = [k for k in required_keys if k not in task]
+        taskstr = task.get("label", task["id"])
+
+        if missing:
+            logger.trace(
+                f"Skipping task '{taskstr}' because it is missing "
+                f"the following attributes: {', '.join(missing)}"
+            )
+            continue
+
+        if task.get("tags"):
+            task["tags"] = {t["name"]: t["value"] for t in task["tags"]}
+
+        if task.get("classification_note"):
+            if isinstance(task["classification_note"], list):
+                task["classification_note"] = task["classification_note"][-1]
+
+        if task.get("_groups"):
+            if isinstance(task["_groups"], str):
+                task["_groups"] = [task["_groups"]]
+
+        if task.get("_result_ok"):
+            oks = task.pop("_result_ok")
+
+            if task.get("_result_group"):
+                groups = task.pop("_result_group")
+
+                task["_results"] = [
+                    GroupResult(group=group, ok=ok) for group, ok in zip(groups, oks)
+                ]
+
+        normalized_tasks.append(task)
+
+    return [Task.create(**task) for task in normalized_tasks]
+
+
 class Push:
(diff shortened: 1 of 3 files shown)


marco-c:
> How can I determine if anything is being cached?

Programmatically, by doing something like `adr.config.cache.has`.

You should also be able to see a huge runtime difference when getting groups and results for all tasks of a push.

armenzg:
I had to deal with `@memoized_property` by moving the logic into an external function that the property calls. I isolated such change as an individual commit to make it easier to check:
https://github.com/mozilla/mozci/pull/112/commits/a69a529e7c3ac83290500f556bd483e0d1b296ba

Without it the test would fail with `TypeError: 'list' object is not callable`:

```
        push = Push("6e87f52e6eebdf777f975baa407d5c22e9756e80", branch="mozilla-beta")
>       tasks = push.tasks()
E       TypeError: 'list' object is not callable

tests/test_integration.py:126: TypeError
============================================================================================================================================ short test summary info =============================================================================================================================================
FAILED tests/test_integration.py::test_caching_of_tasks - TypeError: 'list' object is not callable
```

marco-c:
> I had to deal with @memoized_property by moving the logic into an external function that the property calls

What if we used @lru_cache instead?
```
    @property
    @lru_cache(maxsize=None)
    def tasks(self):
```

tests/test_integration.py
```diff
@@ -89,10 +91,48 @@ def test_good_result_manifests():
...
         if group is None:
             continue
 
         assert (
             not task.is_bad_group("x", group) and "\\" not in group
         ), f"{group} group is bad!"
+
+
+def test_caching_of_tasks(tmp_path):
+    config_file = tmp_path / "config.toml"
```
marco-c, line 104: Nit: you could make the adr config a fixture so other tests in the future can make use of it

tests/test_integration.py
```diff
@@ -89,10 +91,48 @@ def test_good_result_manifests():
...
+retention = 1000
+[adr.cache.stores]
+file = { driver = "file", path = "%s" }
+"""
+        % tmp_path
+    )
+    config_file.write_text(text)
+    config = Configuration(path=config_file)
+
+    # Once we reach Nov. 23rd, 2020 the test should
```
marco-c, line 120: The sentence is cut :P

tests/test_integration.py
```diff
@@ -89,10 +91,48 @@ def test_good_result_manifests():
...
+    for group, _ in result["data"]:
         if group is None:
             continue
 
         assert (
             not task.is_bad_group("x", group) and "\\" not in group
         ), f"{group} group is bad!"
+
+
+def test_caching_of_tasks(tmp_path):
```
marco-c, line 103: Nit: You could also write a unit test, by mocking the answers from Taskcluster via responses

tests/test_integration.py
```diff
@@ -89,10 +91,48 @@ def test_good_result_manifests():
...
+
+    # Once we reach Nov. 23rd, 2020 the test should
+    TASK_ID = "WGNh9Xd8RmSG_170-0-mkQ"
+    # Making sure there's nothing left in the cache
+    assert config.cache.get(TASK_ID) is None
+    # Push from Nov. 22nd, 2019
+    push = Push("6e87f52e6eebdf777f975baa407d5c22e9756e80", branch="mozilla-beta")
+    tasks = push.tasks()
+    for t in tasks:
+        if t.id == TASK_ID:
```
marco-c, line 128: I'd make sure we actually do find the ID in the tasks, otherwise if we introduce a regression somewhere the test might still pass

mozci/task.py
```diff
@@ -164,10 +165,35 @@ def update_group(result):
...
+        if self._errors or self._groups or self._results:
+            logger.debug("Storing {} in the cache".format(self.id))
+            adr.config.cache.put(
+                self.id,
+                {
+                    "errors": self._errors,
+                    "groups": self._groups,
+                    "results": self._results,
+                },
+                1,  # XXX: What should the value in minutes be?
```
marco-c, line 181: We could use the same value as the one in the ADR configuration

mozci/task.py
```diff
@@ -164,10 +165,35 @@ def update_group(result):
...
+            )
+
     def _load_errorsummary(self):
+        # I don't believe we need to invalidate the data since once we parse
+        # the error summary artifact we should not expect a different result in
+        # the future since the artifact doesn't change
```
marco-c, line 187: The only problem is if we change the format of the properties of the tasks

marco-c: Or e.g. if we add more properties

mozci/task.py
```diff
@@ -164,10 +165,35 @@ def update_group(result):
                 if not is_bad_group(self.id, result.group)
             ]
 
+        # I'm storing in the cache in here rather than at the end of _load_errorsummary()
+        # because there's various modifications of various self properties.
```
marco-c, line 169: This means we will store this data even when we already have it in the cache from AD.
What if we stored it in _load_errorsummary, after the call to __post__init__? Or, you could store the raw values and then call __post__init__ also when the data is coming from the cache

tests/test_integration.py
armenzg: Do you have any trick on how to collect all the responses?
One thing I've always struggled with mocking is collecting all the URLs and the individual responses.

mozci/task.py
armenzg: Do you have any suggestions as to how or when to invalidate the data?
A function that clears the cache for all the tasks associated to a push?

mozci/task.py
armenzg: I've adjusted the code to put it after the **post_Iinit** call. Thanks!

What should we do when there's data from AD?

mozci/push.py
```diff
@@ -213,129 +329,13 @@ def decision_task(self):
         task_id = find_task_id(index)
         return Task(id=task_id)
 
-    @memoized_property
```
armenzg, line 232: > > I had to deal with @memoized_property by moving the logic into an external function that the property calls
> 
> What if we used @lru_cache instead?
> 
> ```
>     @property
>     @lru_cache(maxsize=None)
>     def tasks(self):
> ```

I tried it and I had the same results

tests/test_integration.py
armenzg: Handled with `assert found_task`.

tests/test_integration.py
armenzg: I believe I handled this.

tests/test_integration.py
armenzg: I decided we can get this landed and see if we want to push for it on a follow up if that's acceptable to you.

mozci/task.py
marco-c: I have no idea :(

marco-c:
> I had to deal with `@memoized_property` by moving the logic into an external function that the property calls. I isolated such change as an individual commit to make it easier to check:
https://github.com/mozilla/mozci/pull/112/commits/a69a529e7c3ac83290500f556bd483e0d1b296ba

> Without it the test would fail with `TypeError: 'list' object is not callable`:

> ```
>         push = Push("6e87f52e6eebdf777f975baa407d5c22e9756e80", branch="mozilla-beta")
> >       tasks = push.tasks()
> E       TypeError: 'list' object is not callable
> 
> tests/test_integration.py:126: TypeError
> ============================================================================================================================================ short test summary info =============================================================================================================================================
> FAILED tests/test_integration.py::test_caching_of_tasks - TypeError: 'list' object is not callable
> ```

Sorry, I hadn't noticed it before, but in your code here it seems like you were trying to call `tasks`, but `tasks` is a property, so you were trying to call a list.
I think you should be able to go back to having a memoized property and replace `tasks = push.tasks()` with `tasks = push.tasks`.

armenzg:
> > I had to deal with `@memoized_property` by moving the logic into an external function that the property calls. I isolated such change as an individual commit to make it easier to check:
> > [a69a529](https://github.com/mozilla/mozci/commit/a69a529e7c3ac83290500f556bd483e0d1b296ba)
> 
> > Without it the test would fail with `TypeError: 'list' object is not callable`:
> 
> > ```
> >         push = Push("6e87f52e6eebdf777f975baa407d5c22e9756e80", branch="mozilla-beta")
> > >       tasks = push.tasks()
> > E       TypeError: 'list' object is not callable
> > 
> > tests/test_integration.py:126: TypeError
> > ============================================================================================================================================ short test summary info =============================================================================================================================================
> > FAILED tests/test_integration.py::test_caching_of_tasks - TypeError: 'list' object is not callable
> > ```
> 
> Sorry, I hadn't noticed it before, but in your code here it seems like you were trying to call `tasks`, but `tasks` is a property, so you were trying to call a list.
> I think you should be able to go back to having a memoized property and replace `tasks = push.tasks()` with `tasks = push.tasks`.

@marco-c Is there a way for Python to let us know what I was accessing it incorrectly?

armenzg:
I moved `@memoized_property` back in place.
Ready to be reviewed.

marco-c:
> @marco-c Is there a way for Python to let us know what I was accessing it incorrectly?

The error message actually told us that (`you're trying to call a list`), but we didn't pay enough attention.

PR #116, opened 2020-03-27
Drop regressions from pushes which were not backed out and had no bustage fix

### LINKED ISSUE #113: Drop regressions from pushes which were not backed out and had no bustage fix
Right now, we penalize regressions when the push has not been backed out. We don't completely drop them because there could be bustage fixes.
If we had a way to detect bustage fixes, we could completely drop them.

We could rely on the following rules to detect, with probably enough precision, if there is a bustage fix:
1) There is a close enough child push where the task/group passes;
2) The child push where the task/group passes was authored by the same author as the push of interest or by a sheriff;
3) The child push where the task/group passes modified one or more files that were also modified by the push of interest.

marco-c:
Fixes #113

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -84,7 +84,7 @@ class Push:
         """The bugs associated with the commits of this push.
 
         Returns:
-            list: A list of bug IDs.
+            set: A set of bug IDs.
         """
         return set(
             sum(
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -87,14 +87,9 @@ class Push:
             set: A set of bug IDs.
         """
         return set(
-            sum(
-                (
-                    bug["no"]
-                    for changeset in self._hgmo.changesets
-                    for bug in changeset["bugs"]
-                ),
-                [],
-            )
+            bug["no"]
+            for changeset in self._hgmo.changesets
+            for bug in changeset["bugs"]
         )
 
     @property
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -470,7 +470,7 @@ class Push:
 
             # Otherwise, if the backout push also contains the backout commit of this push,
             # we can consider it as a regression of this push.
-            for backout in fix_hgmo.backouts.keys():
+            for backout in fix_hgmo.backouts:
                 if backout[:12] == self.backedoutby[:12]:
                     return True
 
@@ -554,7 +554,7 @@ class Push:
             if len(self.bugs & other.bugs) > 0 and any(
                 name in other.label_summaries
                 and other.label_summaries[name].status == Status.PASS
-                for name in self.get_candidate_regressions("label").keys()
+                for name in self.get_candidate_regressions("label")
             ):
                 return other.rev
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -79,6 +79,24 @@ class Push:
         """
         return bool(self.backedoutby)
 
+    @memoized_property
+    def bugs(self):
+        """The bugs associated with the commits of this push.
+
+        Returns:
+            list: A list of bug IDs.
+        """
+        return set(
+            sum(
+                (
+                    bug["no"]
+                    for changeset in self._hgmo.changesets
+                    for bug in changeset["bugs"]
+                ),
+                [],
+            )
+        )
+
     @property
     def date(self):
         """The push date.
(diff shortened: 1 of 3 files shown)


mozci/push.py
```diff
@@ -78,6 +78,24 @@ def backedout(self):
         """
         return bool(self.backedoutby)
 
+    @memoized_property
+    def bugs(self):
+        """The bugs associated with the commits of this push.
+
+        Returns:
+            list: A list of bug IDs.
```
ahal: nit: it returns a set

mozci/push.py
```diff
@@ -78,6 +78,24 @@ def backedout(self):
...
+        """
+        return set(
+            sum(
+                (
+                    bug["no"]
+                    for changeset in self._hgmo.changesets
+                    for bug in changeset["bugs"]
+                ),
+                [],
+            )
```
ahal: This looks really confusing to me. Isn't `bug["no"]` already a string? How are we adding that to a list? Looks like the call to `sum` should be removed.

mozci/push.py
```diff
@@ -471,9 +515,25 @@ def get_regressions(self, runnable_type):
...
+        # If the push was not backed-out and was not "bustage fixed", it can't
+        # have caused regressions.
+        if not self.backedout and not self.bustage_fixed_by:
+            return regressions
+
+        # Any failure that comes after this push has been backed-out or bustage fixed,
+        # can't be blamed on this push.
+        other = self.child
+        for child_count_cutoff in range(MAX_DEPTH + 1):
+            if self.backedoutby in other.revs or self.bustage_fixed_by in other.revs:
```
ahal: Maybe we should have a `fixedby` property that combines both the `backedoutby` and `bustage_fixed_by` properties into one. We could still differentiate backouts via `push.backedout` if we needed to.

mozci/push.py
```diff
@@ -446,14 +464,40 @@ def get_candidate_regressions(self, runnable_type):
...
+        if self.backedout:
+            return False
+
+        count = 0
+        other = self.child
+        while count < MAX_DEPTH:
+            if len(self.bugs & other.bugs) > 0 and any(
+                name in other.label_summaries
+                and other.label_summaries[name].status == Status.PASS
+                for name in self.get_candidate_regressions("label").keys()
```
ahal: Looks like we walk down the childs both here and in `get_candidate_regressions`. Could we try to make sure we only walk down children once?

Maybe a `_process_children` helper that sets everything up the first time it is called and is a no-op on subsequent times. I guess the expensive parts are memoized anyway, so maybe it's not worth it.. Feel free to ignore this for now.

mozci/push.py
marco-c: This is accumulating the bugs from all the changesets in the push in a single set.
It would be equivalent to:
```Python
bugs = set()
for changeset in self._hgmo.changesets:
    for bug in changeset["bugs"]:  # there should be one per commit, but for some reason "bugs" is a list.
        bugs.add(bug)
```

mozci/push.py
marco-c: Oops, fixed in dbc86ab.

mozci/push.py
marco-c: It makes sense to do it at some point anyway, as we could use the `bustage_fixed_by` to check the sheriffs' classifications.

marco-c: To clarify my previous comment: if we don't untangle bustage_fixed_by from get_candidate_regressions, we would create a circular dependency.

I've filed #125.

mozci/push.py
marco-c: I think we can do this as part of #125.

mozci/push.py
marco-c: Would you prefer the more explicit version? It's slower, but of course it doesn't matter at all here.

ahal: I think there's a bug, the call to `sum` fails trying to concatenate a string to a list. Simply removing the `sum` call worked for me. I don't mind the shortened version.

ahal: To iillustrate, your code is equivalent to:
```python
bugs = []
for changeset in self._hgmo.changesets:
    for bug in changeset["bugs"]:  # there should be one per commit, but for some reason "bugs" is a list.
        bugs.append(bug["no"])

bugs = set(sum(bugs, []))
```

marco-c: You're right, I don't know what I was thinking :P

marco-c: Fixed with bf8d477.

PR #117, opened 2020-03-27
Simplify get_regressions by having a single 'count' variable

marco-c:
Other than a cleanup, this has a benefit: if we have a failure on a push with 'fixed by commit' info we add it to the list of regressions of the current push even if it was intermittent in the parent (given that we won't consider it a "prior regression")

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -471,14 +471,13 @@ class Push:
         """
         regressions = {}
 
-        for name, (child_count, status) in self.get_candidate_regressions(
+        for name, (count, status) in self.get_candidate_regressions(
             runnable_type
         ).items():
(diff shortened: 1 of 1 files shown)


marco-c:
This caused a test failure with a test which I introduced in #121. I changed the expectation in https://github.com/mozilla/mozci/commit/69d17de064c2d81fbe14f991a7cb1e14831e9377.
Basically, when two pushes have a 'fixed by commit' classification pointing to the same backout, we can't tell which one caused the regression (they might even be two different regressions!). The test was checking that only the first was blamed, I changed the expecation to blame both. I assume this situation must be pretty uncommon anyway.

PR #121, opened 2020-03-29
Improve the way we handle 'fixed by commit' classifications

### LINKED ISSUE #118: Ignore clearly wrong classifications
The failure of task IX96oXeZQwSBnre9VItdCg on f848fd145194dea1bee4f54660ad6b6aa8c4b6f9 is classified as fixed by commit 89826f85b5bd69574f7e268b67c24487d70d1d95.
89826f85b5bd69574f7e268b67c24487d70d1d95 backs out efb82594b292ffc7e1da93bfa863f0eef5278fa5.

efb82594b292ffc7e1da93bfa863f0eef5278fa5 landed **after** f848fd145194dea1bee4f54660ad6b6aa8c4b6f9, so it clearly can't be the cause of a regression occurring on f848fd145194dea1bee4f54660ad6b6aa8c4b6f9.

Similarly with the Wr1 failure on cdf202c2682eccdb9781145451cb707ec157db9d. It is marked as fixed by commit aca7870ec9360ec8062b3300aea1dfd70a2b1652, which is a back-out of 1fa5a28777468538ce28c0b8c9a2ac69c391ba3c.
1fa5a28777468538ce28c0b8c9a2ac69c391ba3c landed **after** cdf202c2682eccdb9781145451cb707ec157db9d, so it can't be the cause of the failure.

NOTE: Perhaps Treeherder could disallow these kind of classifications.

marco-c:
In particular, ignore classifications which are clearly wrong.
Fixes #118

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -10,6 +10,7 @@ from adr.query import run_query
 from adr.util.memoize import memoized_property
 from loguru import logger
 
+from mozci.errors import PushNotFound
 from mozci.task import GroupResult, GroupSummary, LabelSummary, Status, Task, TestTask
 from mozci.util.hgmo import HGMO
 from mozci.util.taskcluster import find_task_id
@@ -412,7 +413,13 @@ class Push:
         #   is landed after the push where the failure occurs (so, it can't have caused it);
         # - the backout push also contains a commit backing out one of the commits of this push.
         for classification_note in fixed_by_commit_classification_notes:
-            fix_hgmo = HGMO.create(classification_note, branch=self.branch)
+            try:
+                fix_hgmo = HGMO.create(classification_note, branch=self.branch)
+            except PushNotFound:
+                logger.warning(
+                    "Classification note ({classification_note}) references a revision which does not exist on push {failure_push.rev}"
+                )
+                return None
             if len(fix_hgmo.backouts) == 0:
                 continue
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -376,7 +376,7 @@ class Push:
 
         return int(duration / 3600)
 
-    def _check_classification(self, other, classifications):
+    def _is_classified_as_cause(self, other, classifications):
         """Checks a 'fixed by commit' classification to figure out what push it references.
 
         Returns:
@@ -494,12 +494,12 @@ class Push:
                     passing_runnables.add(name)
                     continue
 
-                classification = self._check_classification(
+                is_classified_as_cause = self._is_classified_as_cause(
                     other, summary.classifications
                 )
-                if classification is True:
+                if is_classified_as_cause is True:
                     candidate_regressions[name] = (-math.inf, summary.status)
-                elif classification is False:
+                elif is_classified_as_cause is False:
                     passing_runnables.add(name)
                     continue
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -375,6 +375,82 @@ class Push:
 
         return int(duration / 3600)
 
+    def _check_classification(self, failure_push, classifications):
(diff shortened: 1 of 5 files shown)


mozci/push.py
```diff
@@ -375,6 +375,82 @@ def scheduled_duration(self):
...
+        # is a backout of another push, it is definitely not a regression of the
+        # current push.
+        # Unless some condition holds which makes us doubt about the correctness of the
+        # classification.
+        # - the backout commit also backs out one of the commits of this push;
+        # - the other push backed-out by the commit which is mentioned in the classification
+        #   is landed after the push where the failure occurs (so, it can't have caused it);
+        # - the backout push also contains a commit backing out one of the commits of this push.
+        for classification_note in fixed_by_commit_classification_notes:
+            fix_hgmo = HGMO.create(classification_note, branch=self.branch)
```
ahal: We should probably catch `PushNotFound` and return `None` here in case there is a typo in the revision.

mozci/push.py
```diff
@@ -375,6 +375,82 @@ def scheduled_duration(self):
 
         return int(duration / 3600)
 
+    def _check_classification(self, failure_push, classifications):
```
ahal: Calling this `failure_push` confused me at first because it implies it is the push that caused the failure, when in reality, that is the information we're trying to discover. I'd prefer we simply call it `other` here as well.

ahal: Is it worth memoizing this?

mozci/push.py
```diff
@@ -411,32 +487,14 @@ def get_candidate_regressions(self, runnable_type):
...
-                    elif all(
-                        HGMO.create(n, branch=self.branch).is_backout
-                        for n in fixed_by_commit_classification_notes
-                    ):
-                        # If the failure was classified as fixed by commit, and the fixing commit
-                        # is a backout of another push, it is definitely not a regression of the
-                        # current push.
-                        passing_runnables.add(name)
-                        continue
+                classification = self._check_classification(
```
ahal: nit: let's call this variable `at_fault` or something similar.

tests/test_push.py
```diff
@@ -185,13 +185,11 @@ def test_succeeded_in_parent_failed_in_current_succeeded_in_child_succeeded_in_g
     assert p[i + 2].get_regressions("label") == {}
 
 
-def test_failure_on_backout(monkeypatch, create_pushes):
+def test_failure_on_backout(create_pushes):
     """
     Tests the scenario where a task succeeded in a parent push, didn't run in the
     push of interest and failed in the push containing the backout of the push of interest.
     """
-    monkeypatch.setattr(HGMO, "is_backout", property(lambda cls: True))
```
ahal: Sorry, I'm just missing it, but where is `backouts` being populated now?

marco-c: I think I had copied this monkeypatching in almost all tests, even when it was not necessary. It's actually needed only in tests which have 'fixed by commit' classifications, as for the other cases we just rely on `backedoutby`.

mozci/push.py
marco-c: Good idea, and I've seen this a few times on a few past pushes. Fixed with ff4c978 (I'm still printing a warning).

mozci/push.py
marco-c: Sorry about that, I couldn't think of a better name! Renamed to `other` in 1df22fa. I don't like the name so much, but at least it's consistent with get_candidate_regressions.

marco-c: Probably not, as we will never call this twice with the same parameters given that we already memoize get_candidate_regressions

mozci/push.py
marco-c: I've renamed both the variable and the function in 32dbc9d.

PR #130, opened 2020-04-01
Refactor bustage_fixed_by and get_candidate_regressions

marco-c:
To avoid dependencies between them, and factor out the duplicated code to iterate children.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,7 +2,7 @@
 import pytest
 from responses import RequestsMock
 
-from mozci.push import Push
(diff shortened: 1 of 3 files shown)


marco-c:
This PR also contains some improvements in tests and fixtures and some performance improvements.
I'd suggest reviewing commit by commit (they are small and self-isolated).

tests/conftest.py
```diff
@@ -86,17 +86,18 @@ def create_pushes(create_push):
```
ahal, line 86: This docstring should be updated.

PR #132, opened 2020-04-01
Fix integration tests

armenzg:


ahal:
The fact that `reftest` flavors are missing manifests is either a legit regression or is exposing some deeper flaw in our test. Either way we should investigate further.

marco-c:
I see the test was skipped in the latest cron run: https://travis-ci.org/github/mozilla/mozci/builds/669726057.

marco-c:
The last build where we run them was https://travis-ci.org/github/mozilla/mozci/builds/663523324, the first where we didn't was https://travis-ci.org/github/mozilla/mozci/builds/663967977. So #91 didn't work as expected.

tests/test_integration.py
```diff
@@ -17,15 +17,20 @@ def test_missing_manifests():
...
+        "reftest",
         "talos",
+        "test-verify-wpt",
+        "web-platform-tests",
+        "web-platform-tests-crashtest",
+        "web-platform-tests-crashtests",
+        "web-platform-tests-reftest",
         "web-platform-tests-reftests",
         "web-platform-tests-wdspec",
-        "web-platform-tests-crashtests",
```
marco-c, line 24: The test passes for me by just adding `web-platform-tests-crashtest` and `web-platform-tests-reftest`

armenzg:
Do you want me to reduce the list and get that landed? or are you still investigating as to why this started happening?

marco-c:
> Do you want me to reduce the list and get that landed?

Yep!

>  or are you still investigating as to why this started happening?

I think it was changed on purpose by @ahal .

tests/test_integration.py
marco-c: Is it passing this way? Or are there both `web-platform-tests-reftests` and `web-platform-tests-reftest`?

armenzg: It seems that only the singular ones are the ones passing.

ahal: This makes sense, Edwin and I recently changed the labels to remove the plural (to match what the WPT harness actually calls them). Sorry I didn't think of this sooner.. for some reason I thought this was happening to normal reftest/crashtest.

marco-c: Yeah I was worried there could still be some of the old ones in the results from the query, but I guess the change has happened enough time ago.

PR #138, opened 2020-04-02
Add ability to find parent pushes on Try

ahal:


tests/test_push.py
```diff
@@ -3,1045 +3,10 @@
 import pytest
```
marco-c, line 3: git is much worse than mercurial here, it doesn't detect the file as moved and edited...
Could you make a first commit where you just rename it, and then a second commit where you create tests/test_push.py extracting the few lines from tests/test_regressions.py? This way blaming won't be painful in the future.

PR #141, opened 2020-04-06
No child

ahal:
Handle the case when a Push has fewer children than `MAX_DEPTH`.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -547,7 +547,11 @@ class Push:
         other = self
         for _ in range(MAX_DEPTH + 1):
             yield other
-            other = other.child
+
+            try:
+                other = other.child
+            except ChildPushNotFound:
+                break
 
     def _iterate_failures(self, runnable_type):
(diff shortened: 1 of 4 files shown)


mozci/push.py
```diff
@@ -547,7 +547,11 @@ def _iterate_children(self):
         other = self
         for _ in range(MAX_DEPTH + 1):
             yield other
-            other = other.child
+
+            try:
+                other = other.child
+            except ChildPushNotFound:
+                break
```
ahal, line 554: I'm kind of assuming that this won't mess up the algorithm at all. Can you think of any ways that this might cause `get_regressions` to fail? Other than making it less accurate..

marco-c: I think it should be OK, we just need to make sure consumers don't cache (or store) the results of recent pushes (since children pushes might appear later).
It's OK in mozci's usage in bugbug since I'm always getting pushes up to three days before when I run the collection job.

PR #142, opened 2020-04-06
Add an API to generate all shadow scheduler results

ahal:
Implements ``Push.generate_all_shadow_scheduler_tasks()``.


ahal:
This should allow the ETL to call this API and retrieve all the shadow-schedulers that ran on a push. My hope is that we can set the ETL up such that new or renamed shadow-schedulers automatically get reflected in the dashboard.

PR #156, opened 2020-04-15
First part for supporting WPT groups

### LINKED ISSUE #153: Support WPT groups
Now that WPT groups are being propagated to ActiveData, mozci needs to handle them.

This involves:
* Updating the integration tests
* Updating the `is_bad_manifest` check in task.py
* ???

marco-c:
First part for #153.

Need to wait a few days before being able to update `test_missing_manifests` and `test_missing_result_manifests`, and there is some issue in ActiveData that @ekyle is looking into.

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -44,12 +44,17 @@ def is_no_groups_suite(label):
 bad_group_warned = False
 
 
-def is_bad_group(task_id, group):
+def is_bad_group(task_id, task_label, group):
     global bad_group_warned
 
-    bad_group = (
-        os.path.isabs(group) or group.startswith("file://") or group.startswith("Z:")
-    )
+    bad_group = group.startswith("file://") or group.startswith("Z:")
+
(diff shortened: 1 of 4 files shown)


mozci/task.py
```diff
@@ -44,12 +44,17 @@ def is_no_groups_suite(label):
...
-def is_bad_group(task_id, group):
+def is_bad_group(task_id, task_label, group):
     global bad_group_warned
 
-    bad_group = (
-        os.path.isabs(group) or group.startswith("file://") or group.startswith("Z:")
-    )
+    bad_group = group.startswith("file://") or group.startswith("Z:")
+
+    # web-platform-tests manifests are currently absolute.
```
marco-c, line 52: It'd be great if we could have web-platform-tests in the same style as other manifests (full relative path in the mozilla-central repository) to avoid special conditions here and in bugbug.

marco-c:
Landing to unblock @ekyle.

PR #157, opened 2020-04-16
Support WPT groups

marco-c:
Second part for #153.


marco-c:
`test_missing_result_manifests` is still failing, but it will start passing once the problem in AD is fixed.

marco-c:
Tests are passing now, since I started ignoring SKIPPED groupts.

PR #159, opened 2020-04-17
Cache tasks' errors/results grouped as a push

### LINKED ISSUE #158: Add coarse cache for task groups and results
Right now, with 7e81abe91b6355086520910d0b8c44e7921e1efe, each task is stored in a different cache key, which makes the number of elements in the cache explode.
We should instead cache all the tasks of a push together.
@armenzg could you look into this?

armenzg:
Fixes #158

The original code to cache tasks was storing each task in a different cache key, which
makes the number of elements in the cache explode. We should instead cache all the tasks
of a push together.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -326,9 +326,11 @@ class Push:
                 )
             )
 
+        # Assign it so it can be used inside of get_groups_for_tasks()
+        self.tasks = tasks
         # Let's gather any new data missing
-        tasks = self.gather_missing_results(tasks)
-        # New data might have been gathered via gather_missing_results; cache it
+        self.get_groups_for_tasks()
+        # New data might have been gathered via get_groups_for_tasks; cache it
         self._cache_test_tasks(tasks)
         return tasks
 
@@ -374,21 +376,6 @@ class Push:
             normalized_tasks.append(task)
 
         return [Task.create(**task) for task in normalized_tasks]
-
-    def gather_missing_results(self, tasks: list) -> list:
-        """ It gathers errors/results from Taskcluster if missing."""
-        # Calling task.results modifies the properties of the task, thus, the returning list
-        # of this function comes back modified
-        future_to_task = {
-            Push.THREAD_POOL_EXECUTOR.submit(lambda task: task.results, task): task
-            for task in tasks
-            if isinstance(task, TestTask)
-        }
-
-        for future in concurrent.futures.as_completed(future_to_task):
-            future_to_task[future]
-
-        return tasks
 
     def _cache_test_tasks(self, tasks: list) -> None:
         test_tasks = {}
@@ -451,6 +438,8 @@ class Push:
         return self.task_labels - self.scheduled_task_labels
 
     def get_groups_for_tasks(self):
+        # Calling task.groups modifies the properties of the task, thus,
+        # the returning list of this function comes back modified
         future_to_task = {
             Push.THREAD_POOL_EXECUTOR.submit(lambda task: task.groups, task): task
             for task in self.tasks
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -318,6 +318,7 @@ class Push:
         # Normalize and validate.
         normalized_tasks = []
         for task in tasks.values():
+            task["push_uuid"] = self.push_uuid
             missing = [k for k in required_keys if k not in task]
             taskstr = task.get("label", task["id"])
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -3,7 +3,6 @@ import concurrent.futures
 import math
 from argparse import Namespace
 from collections import defaultdict
-from dataclasses import field
 from functools import lru_cache
 
 import adr
@@ -31,9 +30,6 @@ class Push:
         revs (list): List of revisions of commits in the push (top-most is the first element).
         branch (str): Branch to look on (default: autoland).
     """
-
-    task: list = field(default=None)
-    push_uuid: str = field(default=None)
 
     # static thread pool, to avoid spawning threads too often.
     THREAD_POOL_EXECUTOR = concurrent.futures.ThreadPoolExecutor()
@@ -237,17 +233,12 @@ class Push:
         return Task.create(index=index)
 
     @memoized_property
-    def tasks(self) -> list:
+    def tasks(self):
         """All tasks that ran on the push, including retriggers and backfills.
 
         Returns:
             list: A list of `Task` objects.
         """
-        push_tasks = adr.config.cache.get(self.push_uuid)
-        if push_tasks:
-            logger.info("Fetched tasks for {} from the cache".format(self.push_uuid))
-            return push_tasks
-
         args = Namespace(rev=self.rev, branch=self.branch)
         tasks = defaultdict(dict)
         retries = defaultdict(int)
@@ -354,12 +345,24 @@ class Push:
 
         tasks = [Task.create(**task) for task in normalized_tasks]
         # Pass tasks instead of trying to access self.tasks which would make this recursive
-        self.cache_tasks(tasks)
+        tasks = self.fetch_groups_errors_results(tasks)
         return tasks
 
-    def cache_tasks(self, tasks):
-        """ This function caches all tasks rather than only error summaries"""
-        push_tasks = defaultdict()
+    def fetch_groups_errors_results(self, tasks: list) -> list:
+        """ This adds errors, groups and results to TestTasks"""
+        push_tasks = adr.config.cache.get(self.push_uuid, defaultdict())
+        if len(push_tasks.keys()) > 0:
+            # Let's add error summaries to TestTasks
+            for t in tasks:
+                error_summary = push_tasks.get(t.id)
+                # Only Test tasks have stored error summary information in the cache
+                if error_summary:
+                    t._errors = error_summary["errors"]
+                    t._groups = error_summary["groups"]
+                    t._results = error_summary["results"]
+            logger.info("Fetched tasks for {} from the cache".format(self.push_uuid))
+            return tasks
+
         # Let's first add non TestTasks
         for task in tasks:
             if not isinstance(task, TestTask):
@@ -382,6 +385,7 @@ class Push:
         adr.config.cache.put(
             self.push_uuid, push_tasks, adr.config["cache"]["retention"],
         )
+        return push_tasks.values()
 
     @property
     def task_labels(self):
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -311,8 +311,8 @@ class Push:
         # Normalize and validate.
         normalized_tasks = []
         for task in tasks.values():
-            # We pass the rev as part of the task to grouping them when caching them
-            task["rev"] = self.rev
+            # We pass the branch/rev as part of the task to grouping them when caching them
+            task["push_uuid"] = "{}-{}".format(self.branch, self.rev)
             missing = [k for k in required_keys if k not in task]
             taskstr = task.get("label", task["id"])
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -5,6 +5,7 @@ from argparse import Namespace
 from collections import defaultdict
 from functools import lru_cache
 
+import adr
 from adr.errors import MissingDataError
 from adr.query import run_query
 from adr.util.memoize import memoized_property
@@ -236,6 +237,13 @@ class Push:
         Returns:
             list: A list of `Task` objects.
         """
+        # XXX: When should we invalidate the data?
+        data = adr.config.cache.get(self.rev)
+        if data is not None:
+            logger.debug(
+                "Tasks for push {} were loaded from the cache.".format(self.rev)
+            )
+            return data
 
         args = Namespace(rev=self.rev, branch=self.branch)
         tasks = defaultdict(dict)
@@ -341,6 +349,13 @@ class Push:
 
             normalized_tasks.append(task)
 
+        # Only cache data if there's something to store
+        if normalized_tasks:
+            logger.debug("Storing push {} in the cache".format(self.rev))
+            # cachy's put() overwrites the value in the cache; add() would only add if its empty
+            adr.config.cache.put(
+                self.rev, normalized_tasks, adr.config["cache"]["retention"],
+            )
         return [Task.create(**task) for task in normalized_tasks]
 
     @property
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -345,12 +345,12 @@ class Push:
 
         tasks = [Task.create(**task) for task in normalized_tasks]
         # Pass tasks instead of trying to access self.tasks which would make this recursive
-        tasks = self.fetch_groups_errors_results(tasks)
+        tasks = self.fetch_errors_results(tasks)
         return tasks
 
-    def fetch_groups_errors_results(self, tasks: list) -> list:
-        """ This adds & caches errors, groups and results to TestTasks"""
-        push_tasks = adr.config.cache.get(self.push_uuid, defaultdict())
+    def fetch_errors_results(self, tasks: list) -> list:
+        """ If missing this function adds & caches errors and results to TestTasks"""
+        push_tasks = adr.config.cache.get(self.push_uuid, {})
         if len(push_tasks.keys()) > 0:
             # Let's add cached error summaries to TestTasks
             for t in tasks:
@@ -358,25 +358,35 @@ class Push:
                 # Only Test tasks have stored error summary information in the cache
                 if error_summary:
                     t._errors = error_summary["errors"]
-                    t._groups = error_summary["groups"]
                     t._results = error_summary["results"]
             logger.info("Fetched tasks for {} from the cache".format(self.push_uuid))
         else:
-            # Let's fetch TestTasks error summaries in parallel
+            # Let's gather the collected data via AD so we can cache it
+            for t in tasks:
+                if isinstance(t, TestTask) and t._results:
+                    push_tasks[t.id] = {
+                        "errors": t._errors,
+                        "results": t._results,
+                    }
+
+            # If there are any test tasks that do not have results from AD let's fetch them
+            # If task.results is not initialized it will be fetched via Taskcluster
+            # Calling task.results modifies the properties of the task, thus, the returning list
+            # of this function comes back modified
             future_to_task = {
-                Push.THREAD_POOL_EXECUTOR.submit(lambda task: task.groups, task): task
+                Push.THREAD_POOL_EXECUTOR.submit(lambda task: task.results, task): task
                 for task in tasks
-                # If AD already initialized the _groups field we don't need to cache that data
-                if isinstance(task, TestTask) and task._groups is None
+                if isinstance(task, TestTask) and task._results is None
             }
 
             for future in concurrent.futures.as_completed(future_to_task):
                 task = future_to_task[future]
+                # We assert that the information was not gathered via AD already
+                assert push_tasks.get(task.id) is None
                 # This test task now has the values for errors, groups & results
                 push_tasks[task.id] = {
-                    "errors": task._errors,
-                    "groups": task._groups,
-                    "results": task._results,
+                    "errors": task._errors or [],
+                    "results": task._results or [],
                 }
 
             # Cache data
@@ -385,11 +395,14 @@ class Push:
                     self.push_uuid
                 )
             )
-            # We *only* cache error summaries
+            # We *only* cache errors and results
             # cachy's put() overwrites the value in the cache; add() would only add if its empty
             adr.config.cache.put(
                 self.push_uuid, push_tasks, adr.config["cache"]["retention"],
             )
+            if adr.config.cache.get(self.push_uuid) is None:
+                logger.warning("The cache has failed to store.")
+
         return tasks
 
     @property
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -308,7 +308,7 @@ class Push:
             except MissingDataError:
                 pass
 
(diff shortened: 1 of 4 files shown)


armenzg:
This can be tested with: `poetry run pytest -s -vv tests/test_integration.py -k test_caching_of_push`

mozci/push.py
```diff
@@ -317,6 +317,8 @@ def add(result):
         # Normalize and validate.
         normalized_tasks = []
         for task in tasks.values():
+            # We pass the rev as part of the task to grouping them when caching them
```
ahal, line 320: nit: "to group them when caching"

armenzg:
I'm confused. I saw the two of you set up as reviewers but not comments on the PR.

@marco-c this is ready to be reviewed.

mozci/task.py
```diff
@@ -73,6 +74,7 @@ class Task:
     classification: str = field(default=None)
     classification_note: str = field(default=None)
     tags: Dict = field(default_factory=dict)
+    rev: str = field(default=None)
```
ahal, line 77: I think instead of `rev` tasks should be associated with a `Push` instance since `rev` is not a unique identifier for a push (the same `rev` can appear on many different branches). A unique identifier would be branch + rev or possibly push id (though this involves querying hgmo to get).

mozci/task.py
```diff
@@ -200,13 +202,13 @@ def update_group(result):
...
-        # Temporarily disable caching as it's creating too many entries.
-        data = None  # adr.config.cache.get(self.id)
-        if data:
-            self._errors = data["errors"]
-            self._groups = data["groups"]
-            self._results = data["results"]
+        # XXX: When should we invalidate the data?
+        push_data = adr.config.cache.get(self.rev)
+        # If there's data for the push and if it contains this task
+        if push_data and push_data.get(self.id):
```
ahal, line 208: nit: I suggest something like:
```
task_data = cache.get(self.rev, {}).get(self.id)
if task_data:
    self._errors = task_data["errors"]
    ...
```

mozci/task.py
marco-c: IIRC push ID isn't unique either

mozci/task.py
```diff
@@ -200,13 +202,13 @@ def update_group(result):
             ]
 
     def _load_errorsummary(self):
```
marco-c, line 204: Could we load/store from the cache outside of _load_errorsummary, just once per push? This way we load/store for every task in a push

mozci/task.py
armenzg: I've handled this. Resolving.

mozci/push.py
```diff
@@ -236,6 +239,10 @@ def tasks(self):
         Returns:
             list: A list of `Task` objects.
         """
+        push_tasks = adr.config.cache.get(self.push_uuid)
+        if push_tasks:
+            logger.info("Fetched tasks for {} from the cache".format(self.push_uuid))
+            return push_tasks
```
armenzg, line 245: I've got a bit carried away and I'm caching all Task instances.
I am only concerned I've cached more than I had been asked to. I'm happy to go back to just storing the groups, errors and results.

mozci/push.py
```diff
@@ -53,6 +54,8 @@ def __init__(self, revs, branch="autoland"):
             self.rev = self._hgmo["node"]
 
         self.index = BASE_INDEX.format(branch=self.branch, rev=self.rev)
+        # Unique identifier for a Push across branches
+        self.push_uuid = "{}/{}".format(self.branch, self.rev)
```
armenzg, line 58: This replaces the previous usage of `rev` as a unique identifier.

mozci/push.py
```diff
@@ -351,7 +359,21 @@ def add(result):
...
             normalized_tasks.append(task)
 
-        return [Task.create(**task) for task in normalized_tasks]
+        tasks = [Task.create(**task) for task in normalized_tasks]
+        self.cache_data(tasks)
+        return tasks
+
+    def cache_data(self, tasks):
+        push_tasks = {}
+        for task in tasks:
```
armenzg, line 368: I had considered iterating here over tasks that are TestTask instances and call `_load_errorsummary()`.
It was very slow. We could parallelize it if that is what you prefer.

mozci/task.py
```diff
@@ -169,13 +171,19 @@ class TestTask(Task):
 
     def __post_init__(self):
         if is_no_groups_suite(self.label):
-            assert self._groups is None, f"{self.label} should have no groups"
+            assert (
+                self._groups is None or len(self._groups) == 0
```
armenzg, line 175: This is because on the first pass we turn None to [] and then we cache it.

mozci/task.py
armenzg: I have made another comment above. I don't think I've quite managed to accomplish it. Let me know if the idea of parallelization is something you would like me to pursue.

mozci/push.py
armenzg: An alternative to parallelization that could work is fetching all tasks as a group as I do in [Treeherder](https://github.com/mozilla/treeherder/blob/master/treeherder/etl/management/commands/ingest.py#L95-L109).

mozci/task.py
marco-c: You could load the groups directly in `Push.tasks`, and avoid any access to the cache in this function.
We already have some code to parallelize loading groups for all tasks of a push at https://github.com/mozilla/mozci/blob/ef786901e7a74022db0370b5b2a524900d0ef4df/mozci/push.py#L405-L413 (which is likely going to move to a separate function with https://github.com/mozilla/mozci/pull/163/files).

marco-c: By executing that, you'd trigger loading groups for all tasks, once that's done you can perform the caching.

mozci/push.py
marco-c: There are two things we should take into account:
- is caching the whole task going to result in a too large cache compared to just caching groups, errors and results?
- ADR already caches part of the same data (or the same data, when groups are available in unittest), so we should disable caching for the adr queries in this function, or we should only do our caching when the data is not available in AD and we had to load it from Taskcluster; if we don't do this, we'd cache twice the data for many pushes.

mozci/push.py
marco-c: I mentioned it in another comment, we already have some code in push.py to do that, so you could reuse it

marco-c: The treeherder code is loading all tasks from a group of tasks, but here it wouldn't help: we already have the tasks of the group (which are the same as the tasks of the push).

marco-c: N.B.: a `group` in mozci is a manifest, a `group` in Taskcluster is a group of tasks

marco-c:
> Taking a step back, would it make sense to ignore pushes that require having to hit Taskcluster for this information? Is this necessary for pushes older than 2-4 months? (I forget)

Yeah, AD only has 12 weeks (or less, I don't remember).

armenzg:
> > Taking a step back, would it make sense to ignore pushes that require having to hit Taskcluster for this information? Is this necessary for pushes older than 2-4 months? (I forget)
> 
> Yeah, AD only has 12 weeks (or less, I don't remember).

My apologies. I was trying to ask if we should just ignore any Pushes beyond what AD holds and remove the need to try to support older data. I am concerned that overtime we will have to keep investing energy on supporting data past of what AD holds. If it's absolutely necessary then let's keep investing. I just want to make sure that is worth it.

armenzg:
Removing review request. Tackling feedback given by marco.

armenzg:
@marco-c I believe I've managed to accomplished what we discussed.

Ready for review.

This can be tested with `poetry run pytest -s -vv tests/test_integration.py -k test_caching_`

armenzg:
I took me a while but I believe that a lot of my problems have to do that adr has some weird initializations of cachy. In the tests I get the fixture to configure correctly the store, however, the `adr` import in `push.py` defaults to the null store.

I will sleep on this.

```
(Pdb) adr.config.cache._stores
{'null': <cachy.repository.Repository object at 0x10d53f350>}
```

marco-c:
> I took me a while but I believe that a lot of my problems have to do that adr has some weird initializations of cachy. In the tests I get the fixture to configure correctly the store, however, the adr import in push.py defaults to the null store.

Do you have a adr configuration file with a cache configured?
Here's mine:
```
marco@marco-Dell-XPS ~/Documenti/workspace/mozci (config_group)$ cat ~/.config/adr/config.toml 
[adr.cache]
retention = 40320

[adr.cache.stores]
file = { driver = "renewing-file", path = "/home/marco/Documenti/workspace/bugbug/data/adr_cache" }
```

armenzg:
When running pytest ADR_CONFIG_PATH is set to `tests/config.toml` but with the fixture I create a different one.

Thus, when `mozci/push.py` imports `adr` that will grab the configuration from `tests/config.toml`.

I'm pondering as to what is the right way mozci should indirectly use the caching capacity of adr or even set up the tests.

armenzg:
I want to dump my thoughts here with regards to caching:

- There should be an exception raised if getting and putting is used with the null store
- mozci should perhaps write a file with configuration if no ADR_CONFIG_PATH is defined
- mozci should perhaps use cachy directly without using adr's configuration paths
- mozci should perhaps a global manager class for the cache or each class to have an option `cache_manager` argument
- Modify tests/config.toml to define an `adr_tests_cache` file store. Such directory is deleted before running tests.

armenzg:
I've decided to hard code the file store path into `tests/config.toml` and get rid of my fixture:
https://github.com/mozilla/mozci/pull/159/files#diff-00cd6920017a6641e4cc88fa67834db6R8

This does what I need.

mozci/push.py
```diff
@@ -341,7 +343,83 @@ def add(result):
...
+            # Let's add cached error summaries to TestTasks
+            for t in tasks:
+                error_summary = test_tasks.get(t.id)
+                # Only Test tasks have stored error summary information in the cache
+                if error_summary:
+                    t._errors = error_summary["errors"]
+                    t._results = error_summary["results"]
+            logger.info("Fetched tasks for {} from the cache".format(self.push_uuid))
+        else:
+            # This list will make reference to tasks that don't have the results or errors properties
```
marco-c, line 364: Instead of doing this, I think it would be easier (and a lot less code would be required) to just avoid caching in the ADR query and then always cache ourselves everything.

mozci/push.py
```diff
@@ -341,7 +343,83 @@ def add(result):
...
             normalized_tasks.append(task)
 
-        return [Task.create(**task) for task in normalized_tasks]
+        tasks = [Task.create(**task) for task in normalized_tasks]
+        # Pass tasks instead of trying to access self.tasks which would make this recursive
+        tasks = self.fetch_errors_results(tasks)
+        return tasks
+
+    def fetch_errors_results(self, tasks: list) -> list:
+        """ Fetch errors and results for test tasks. Define ADR_CONFIG_PATH and adr.cache for the caching to work"""
```
marco-c, line 352: I wouldn't put the second sentence in the documentation of the function, it's not really directly related (otherwise, we'd have to put the same sentence in all functions using adr :P)

marco-c:
NOTE: when the data is not available in AD, with the current implementation we'd still perform the query every time `tasks` is accessed, since you perform the query first, and then check the cache later.

mozci/push.py
armenzg: I can add the parameter to avoid the caching (that got lost when the 3rd `run_query` got dropped), however, ADR will fetch that data regardless if we say it not to cache it.

In other words, ADR already got some of the data for us so instead of fetching it again via Taskcluster we simply cache it.

marco-c: Yeah, we should check the cache first and if it's in the cache avoid running the query (which would also fix my other comment https://github.com/mozilla/mozci/pull/159#issuecomment-623461999).

ahal:
There's been a ton of discussion here that I haven't been following along, and I trust Marco's review will be good enough. So removing myself as a reviewer.

armenzg:
Thanks for the catch. I think I'm missing the forest for the trees.

armenzg:
Trying to lay down the flow of the code in here to make sure it makes sense.

Empty cache:

1. Before we try to fetch the data from AD we check the cache (`test_tasks_results = adr.config.cache.get(self.push_uuid, {})`
2. We get the data from AD **without** caching it (`run_query("push_tasks_results_from_unittest", no_cache_args)`
3. We gather missing results and cache both AD & TC data (`self.gather_missing_results(self.normalized_tasks(tasks))`)

Already cached:
1. Before we try to fetch the data from AD we check the cache (`test_tasks_results = adr.config.cache.get(self.push_uuid, {})`
2. For each **test** task we initialize the properties `_errors` and `_results` from the cache

To test the scenario `already cached` you can call the test twice by commenting this out:
- shutil.rmtree("adr_tests_cache")
- Three lines in [here](https://github.com/mozilla/mozci/pull/159/files#diff-7eb47fc18d4a6196f0940784484a5bf3R165)

Steps to run test:
```shell
poetry install
poetry shell
export TRAVIS_EVENT_TYPE=cron
export ADR_CONFIG_PATH=tests/config.toml
pytest tests/test_integration.py -s -vv -k test_caching_
```

mozci/push.py
```diff
@@ -294,14 +296,36 @@ def add(result):
...
+                )
+            )
+            return tasks
+        else:
+            # Gather information from the unittest table. We allow missing data for this table because
+            # ActiveData only holds very recent data in it, but we have fallbacks on Taskcluster
+            # artifacts.
+            try:
+                no_cache_args = Namespace(rev=self.rev, branch=self.branch, cache=False)
+                add(run_query("push_tasks_results_from_unittest", no_cache_args))
```
marco-c, line 323: This doesn't seem right, it should be `add(run_query("push_tasks_results_from_unittest", args, cache=False))`, no?

mozci/push.py
```diff
@@ -294,14 +296,36 @@ def add(result):
...
+            except MissingDataError:
+                pass
+            return self.gather_missing_results(self.normalized_tasks(tasks))
 
+    def normalized_tasks(self, tasks):
```
marco-c, line 328: Nit: This is internal, so I'd rename it `_normalize_tasks` (and it could even be static, since it doesn't use any attribute of Push).

mozci/push.py
```diff
@@ -343,6 +367,68 @@ def add(result):
...
+        for t in tasks:
+            if isinstance(t, TestTask):
+                if not (t._results is None):
+                    # Data for this task has been gathered
+                    test_tasks[t.id] = {
+                        "errors": t._errors,
+                        "results": t._results,
+                    }
+                else:
+                    tasks_missing_results.append(t)
```
marco-c, line 385: You should be able to remove all of this code and just get errors and results for all tasks. If the data is available already, it will be returned, if the data is not available already, it will be gathered from errorsummary.log. There should be no need to first get the tasks with missing results.

mozci/push.py
```diff
@@ -343,6 +367,68 @@ def add(result):
...
+
+        for future in concurrent.futures.as_completed(future_to_task):
+            task = future_to_task[future]
+            logger.debug("Fetching errorsummary for {}".format(task.id))
+            # We assert that the information was not gathered via AD already
+            assert test_tasks.get(task.id) is None
+            # This test task now has the values for errors & results
+            test_tasks[task.id] = {
+                "errors": task.errors,
+                "results": task.results,
```
marco-c, line 406: Can we cache the internal variables, _errors and _results?

mozci/push.py
```diff
@@ -343,6 +367,68 @@ def add(result):
...
+                )
+            )
+            # We *only* cache errors and results
+            # cachy's put() overwrites the value in the cache; add() would only add if its empty
+            adr.config.cache.put(
+                self.push_uuid, test_tasks, adr.config["cache"]["retention"],
+            )
+
+            if adr.config.cache.get(self.push_uuid) is None:
+                logger.warning("The cache has failed to store the data.")
```
marco-c, line 428: Could you remove this? It seems overly cautious and we'd pay the price of getting the data from S3 even if we don't need to.

mozci/push.py
```diff
@@ -343,6 +367,68 @@ def add(result):
...
+            test_tasks[task.id] = {
+                "errors": task.errors,
+                "results": task.results,
+            }
+
+        logger.info(
+            "Obtained {} errors/results from Taskcluster".format(
+                len(tasks_missing_results)
+            )
+        )
```
marco-c, line 413: We'd be logging to much with this, can you either remove it or make it debug level?

mozci/push.py
```diff
@@ -294,14 +296,36 @@ def add(result):
...
 
-        # Gather information from the unittest table. We allow missing data for this table because
-        # ActiveData only holds very recent data in it, but we have fallbacks on Taskcluster
-        # artifacts.
-        try:
-            add(run_query("push_tasks_results_from_unittest", args))
-        except MissingDataError:
-            pass
+        # Let's gather error/results from cache or AD/Taskcluster
+        test_tasks_results = adr.config.cache.get(self.push_uuid, {})
```
marco-c, line 300: I'm thinking an implementation like this would be cleaner and the diff would be much smaller:
```
test_tasks_results = adr.config.cache.get(self.push_uuid, {})
was_cached = len(test_tasks_results.keys()) != 0
if not was_cached:
    # get from AD

# normalize tasks

# "extract" test_tasks_results

# load errors and results for all tasks

if not was_cached:
   # store to cache
```

WDYT?

mozci/push.py
```diff
@@ -294,14 +296,36 @@ def add(result):
...
+                    error_summary = test_tasks_results.get(t.id)
+                    # Only Test tasks have stored error summary information in the cache
+                    if error_summary:
+                        t._errors = error_summary["errors"]
+                        t._results = error_summary["results"]
+            logger.info(
+                "Fetched tasks errors/results for {} from the cache".format(
+                    self.push_uuid
+                )
+            )
```
marco-c, line 315: Same here, either remove or make it debug (it's not really information a consumer of the library is interested in).

mozci/push.py
armenzg: You're right!

mozci/push.py
armenzg: Cool! That makes sense. Changing it now.

mozci/push.py
armenzg: I will.

mozci/push.py
armenzg: Let me know what you think of the refactoring.

armenzg:
I've addressed your feedback. Please let me know.

mozci/task.py
```diff
@@ -265,35 +258,22 @@ def _load_errorsummary(self):
...
-            #         "errors": self._errors,
-            #         "results": self._results,
-            #     },
-            #     adr.config["cache"]["retention"],
-            # )
 
     @property
     def groups(self):
-        if self._results is None:
+        if not self._results:
```
marco-c, line 264: If _results (or _errors in errors) is a list with 0 elements, we'd be calling _load_errorsummary again. I don't remember exactly why you had to do this in a previous iteration of the patch, but maybe it's no longer needed now?

mozci/push.py
```diff
@@ -343,6 +375,47 @@ def add(result):
...
+            Push.THREAD_POOL_EXECUTOR.submit(lambda task: task.results, task): task
+            for task in tasks
+            if isinstance(task, TestTask)
+        }
+
+        # XXX: Using threading even for tasks that have data in AD can be slow. This push we're testing
+        # has more than 95% of all tasks in AD. If in test_integration you comment out lines 26 and
+        # lines 165-167 you will see that it takes 4 seconds with caching (it was almost instant in the previous iteration)
+        # and 6 seconds when there's no caching (This means fetching the TC data takes 2 seconds; about 70 tasks).
+        # I'm simply calling out FYI but I think it is fine
```
marco-c, line 392: I wonder why, the threads for the tasks that already have data from AD should be super-fast... Anyway, yeah, I guess it's fine

mozci/task.py
armenzg: I got confused with seeing `Fetching errorsummary for` for every single task.

armenzg:
Addressed.

You can find the constant tasks that always requires hitting Taskcluster (`t._results is None`):
https://gist.github.com/armenzg/e4fa3ae6a572224385427a4b7da54080

mozci/push.py
```diff
@@ -294,14 +296,44 @@ def add(result):
...
+                        t._errors = error_summary["errors"]
+                        t._results = error_summary["results"]
+            logger.debug(
+                "Fetched tasks errors/results for {} from the cache".format(
+                    self.push_uuid
+                )
+            )
 
+        # Let's gather any new data missing
+        tasks = self.gather_missing_results(tasks)
```
marco-c, line 330: Nit: since we have already another function in Push which does almost the same, could you remove this and just use `get_groups_for_tasks`?
```
# Load errors and results for all tasks.
collections.deque(self.get_groups_for_tasks(), maxlen=1)
```

armenzg:
I've addressed the last request.
I will squash it and merge it once Travis comes back passing.

Thank you Marco for your endless reviews!

mozci/push.py
marco-c: Oops, you called get_groups_for_tasks but didn't consume it! I'll fix it in a follow-up commit.

armenzg:
Follow up code landed here:
https://github.com/mozilla/mozci/commit/418b20339d14415adbbeee6619dd397a12ccb965

mozci/push.py
armenzg: The follow up code refered to:
https://github.com/mozilla/mozci/commit/418b20339d14415adbbeee6619dd397a12ccb965

PR #163, opened 2020-04-20
Support getting regressions for groups at given configurations

### LINKED ISSUE #162: Support analyzing pushes at the (platform/configuration, manifest) level
We currently consider two types of runnables: label and group.
For the group runnable, we bundle together runs of the same group on multiple configurations.

In order to make bugbug select the platform/config too (https://github.com/mozilla/bugbug/issues/1117), we should create another type of runnable which takes into account both the platform/config AND the group.

NOTE: if a push runs a group on a given platform but doesn't run the same group on another platform, we can't be sure if it would have failed on that other platform too or not. This makes checking if a regression is a prior regression a bit less precise.
E.g., say group A fails on Windows in push 1 and fails on Linux in push 2. At the group level, it would be considered as a prior regression. At the (platform/config, group) level, it would not be considered as a prior regression.

marco-c:
Fixes #162

Example output for `Push("a2bff0eef797dc2f2f40779eb08ccfae0ceeb27f").get_regressions("group_config")`:
```
{
	('windows7-32', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
	('macosx1014-64', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
	('linux1804-64-shippable', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('linux1804-64-qr', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('macosx1014-64-shippable', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('windows10-64', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('linux1804-64', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('windows10-64-shippable', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('windows10-64-shippable', 'toolkit/components/extensions/test/xpcshell/xpcshell-common.ini'): 0, ('windows7-32-shippable', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('linux1804-64-asan', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('linux1804-64-shippable-qr', 'extensions/permissions/test/unit/xpcshell.ini'): 0,
        ('linux1804-64-tsan', 'extensions/permissions/test/unit/xpcshell.ini'): 0
}
```

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -383,15 +383,7 @@ class Push:
         """
         return self.task_labels - self.scheduled_task_labels
 
-    @memoized_property
-    def group_summaries(self):
-        """All group summaries combining retriggers.
-
-        Returns:
-            dict: A dictionary of the form {<group>: [<GroupSummary>]}.
-        """
-        groups = defaultdict(list)
-
+    def get_groups_for_tasks(self):
         future_to_task = {
             Push.THREAD_POOL_EXECUTOR.submit(lambda task: task.groups, task): task
             for task in self.tasks
@@ -401,10 +393,38 @@ class Push:
         for future in concurrent.futures.as_completed(future_to_task):
             task = future_to_task[future]
             for group in future.result():
-                groups[group].append(task)
+                yield task, group
 
-        groups = {group: GroupSummary(group, tasks) for group, tasks in groups.items()}
-        return groups
+    @memoized_property
+    def group_config_summaries(self):
+        """All group summaries, on given configurations, combining retriggers.
+
+        Returns:
+            dict: A dictionary of the form {(<configuration>, <group>): [<GroupSummary>]}.
+        """
+        config_groups = defaultdict(list)
+
+        for task, group in self.get_groups_for_tasks():
+            config_groups[(task.configuration, group)].append(task)
+
+        return {
+            config_group: GroupSummary(config_group[1], tasks)
+            for config_group, tasks in config_groups.items()
+        }
+
+    @memoized_property
+    def group_summaries(self):
+        """All group summaries combining retriggers.
+
+        Returns:
+            dict: A dictionary of the form {<group>: [<GroupSummary>]}.
+        """
+        groups = defaultdict(list)
+
+        for task, group in self.get_groups_for_tasks():
+            groups[group].append(task)
+
+        return {group: GroupSummary(group, tasks) for group, tasks in groups.items()}
 
     @memoized_property
     def label_summaries(self):
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -396,7 +396,7 @@ class Push:
                 yield task, group
 
     @memoized_property
-    def group_config_summaries(self):
+    def config_group_summaries(self):
         """All group summaries, on given configurations, combining retriggers.
 
         Returns:
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -22,6 +22,39 @@ class Status(Enum):
     INTERMITTENT = 2
 
 
+SUITES = (
+    "mochitest-plain-gpu",
+    "mochitest-plain",
+    "mochitest-chrome-gpu",
+    "mochitest-chrome",
+    "mochitest-devtools-chrome",
+    "mochitest-browser-chrome",
+    "web-platform-tests-wdspec",
+    "web-platform-tests",
+    "mochitest-media",
+    "mochitest-webgpu",
+    "mochitest-webgl1-ext",
+    "mochitest-webgl2-ext",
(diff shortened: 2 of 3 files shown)


mozci/task.py
```diff
@@ -253,6 +253,10 @@ def _load_errorsummary(self):
             #     adr.config["cache"]["retention"],
             # )
 
+    @property
+    def configuration(self):
+        return self.label[len("test-") : self.label.index("/")]
```
marco-c, line 258: I wonder if there is a better way to get this

mozci/push.py
```diff
@@ -411,10 +403,38 @@ def group_summaries(self):
...
+                yield task, group
+
+    @memoized_property
+    def group_config_summaries(self):
+        """All group summaries, on given configurations, combining retriggers.
+
+        Returns:
+            dict: A dictionary of the form {(<configuration>, <group>): [<GroupSummary>]}.
+        """
+        config_groups = defaultdict(list)
```
marco-c, line 415: I wish I could find a better term to define a (<configuration>, <group>) couple...

ahal: Wouldn't that be a `runnable`?

mozci/task.py
ahal: Yeah, I'm not sure if this will work since certain things (like `-fis`) appear after the `/`. The build type (debug/opt) also comes after the slash. I'm not sure the best way of extracting this.. Maybe a regex?

To do it properly likely involves loading the actual task definition from the taskgraph and inspecting attributes, but that sounds expensive.

ahal: I'm kind of assuming you care about `build_type` and `unittest_variants` (aka fis, spi, etc) here.

mozci/push.py
marco-c: Yeah but by `runnable` we basically mean "a possible grouping of task", so "label", "group" and this new one are all possible "runnables". I'd like a name which is not exactly "group" (as we still need that until bugbug uses this "configgroup" only).

mozci/task.py
marco-c: Yeah, I care about any alternative platform/config/whatever where a group could run

marco-c: Maybe the best way could be to get the full name and remove suite names and numbers?

E.g. `test-linux1804-64-qr/debug-mochitest-plain-fis-e10s-7` -> `test-linux1804-64-qr/debug-*-fis-e10s-*`

marco-c: Another option would be to get those attributes as part of the AD queries that we already do, if the data is available in the Treeherder or Unittest table (e.g. in unittest I see build.type and run.type could be useful).

In summary, I need a way to have something that I will return in the bugbug reply and that the taskgraph can easily use to select the platform/configuration.

ahal: Yeah removing the suite from the label seems like a good bet for now. But having this "unique config" value easily accessible seems useful for other things. It might be worth investigating storing it somewhere accessible in the taskgraph such that it gets picked up by AD.

mozci/push.py
ahal: Maybe we should just expand the meaning of `test_platform` to include runtime information such as `fission`. Another alternative might be `test_context`, though `context` is also used frequently.

mozci/task.py
```diff
@@ -264,6 +264,51 @@ def errors(self):
             self._load_errorsummary()
         return self._errors
 
+    @property
+    def configuration(self):
+        SUITES = (
```
ahal, line 269: Can you move this to a global? Possibly even in a utility file?

mozci/task.py
```diff
@@ -264,6 +264,51 @@ def errors(self):
...
+
+        config = self.label
+        for s in SUITES:
+            if f"-{s}-" in config:
+                config = config.replace(s, "*")
+
+        parts = config.split("/")
+
+        return "{}/{}".format(
+            parts[0], "-".join(p for p in parts[1].split("-") if not p.isdigit())
```
ahal, line 309: Are there cases where the chunk number doesn't come at the end? If not, I think we can rely on it being at the end and this seems overly complicated.

mozci/task.py
marco-c: Done with b2f644e.

mozci/task.py
marco-c: I wasn't sure, I've checked on a backstop push and it seems it's always the case, so I've simplified this in f13f716.

PR #180, opened 2020-04-29
Test failure to cache

armenzg:


tests/test_cache.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_cache_with_improper_config(tmp_path):
+    contents = """
+[adr]
+verbose = 1
+"""
+    config_path = tmp_path / "config.toml"
+    cache = obtain_cache(config_path, contents)
+    cache.put("foo", "bar", 5)
+    assert cache.get("foo") == "bar"
```
armenzg, line 39: This is the issue I've been facing all along. Unless the contents of config.toml are correct we can be using the Null store.

Should I change some code on Adr to prevent using the cache property unless the default store is different than the null store?

armenzg: Failing tests https://travis-ci.org/github/mozilla/mozci/builds/681058159

marco-c:
I guess we can close this as it served its purpose.
@armenzg could you file an issue in the adr repo about logging a warning when a cache backend is not selected?

armenzg:
Filed. Thanks!
https://github.com/mozilla/adr/issues/103

PR #182, opened 2020-04-30
Transform WPT group names to a full relative path in mozilla-central

### LINKED ISSUE #161: Make WPT return full relative paths and adjust `is_bad_group` check
Currently, with 61720ebc59424a955458ad2be43560b2e4401ddd, we have a special case for WPT tasks (https://github.com/mozilla/mozci/blob/11777df4ad7b8d21922c37b8a51c80b5f1fc690f/mozci/task.py#L50-L55).

We should check if the harness could return full relative paths, so we have the same behavior as the other suites.

marco-c:
Fixes #161

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -75,23 +75,29 @@ def is_no_groups_suite(label):
 bad_group_warned = False
 
 
-def is_bad_group(task_id, task_label, group):
+def is_bad_group(task_id: str, group: str) -> bool:
     global bad_group_warned
 
-    bad_group = group.startswith("file://") or group.startswith("Z:")
-
-    # web-platform-tests manifests are currently absolute.
-    if not any(
-        sublabel not in task_label
-        for sublabel in {"web-platform-tests", "test-verify-wpt"}
-    ):
-        bad_group |= os.path.isabs(group)
+    bad_group = (
+        group.startswith("file://") or group.startswith("Z:") or os.path.isabs(group)
+    )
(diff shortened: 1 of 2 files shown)


mozci/task.py
```diff
@@ -206,19 +216,27 @@ def __post_init__(self):
...
+            return
+
         # XXX: A while after bug 1613937 and bug 1613939 have been fixed, we can
         # remove the filtering and slash replacing.
 
-        def update_group(result):
+        # Apply WPT workaround, needed at least until bug 1632546 is fixed.
+        if self.is_wpt:
+            for result in self._results:
+                result.group = wpt_workaround(result.group)
```
ahal, line 228: It looks like this means we only apply the workaround if we pass in the results from the constructor (i.e data was loaded from AD). Could you refactor this so it also gets applied when loaded via the errorsummary?

I realize that we already had this problem to an extent with the replacing of slashes. So feel free to file a follow-up issue if you prefer.

marco-c: We are actually calling the `__post_init__` function manually in `_load_errorsummary` (it's a bit of a hack, I know). I can move the clean-up code in a separate function if you want and call it both from `__post_init__` and `_load_errorsummary` (removing the `__post_init__` call from `load_errorsummary`).

marco-c: I'll just merge as-is, we can clean up later if we want.

PR #188, opened 2020-05-06
Ignore skipped groups (from errorsummary.log)

### LINKED ISSUE #166: Ignore skipped groups
Skipped groups are currently considered as passing groups, we should instead ignore them.

marco-c:
Fixes #166

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -243,7 +243,7 @@ class TestTask(Task):
         for result in self._results:
             result.group = result.group.replace("\\", "/")
 
-    def _load_errorsummary(self):
+    def _load_errorsummary(self) -> None:
         # This may clobber the values that were populated by ActiveData, but
         # since the artifact is already downloaded, parsed and we need to
         # iterate over it anyway. It doesn't really hurt and simplifies some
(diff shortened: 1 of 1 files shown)


mozci/task.py
```diff
@@ -261,34 +261,56 @@ def _load_errorsummary(self):
...
             # Assume all groups for which we have no results passed.
-            # In practice, they could also have been skipped (we should ignore them when
-            # it will be feasible, https://github.com/mozilla/mozci/issues/166).
-            known_results = {result.group for result in self._results}
+            # After enough time has passed since https://bugzilla.mozilla.org/show_bug.cgi?id=1632242, we
+            # can stop assuming as all groups will have an associated result.
+            if has_group_result:
+                assert set(groups) == set(
+                    group_results
+                ), f"There are some groups with no results in task {self.id}"
```
marco-c, line 308: I'm not sure why, if I inject an `assert False` here, the exception is eaten somewhere along the chain (I confirmed the line was hit by adding print statements before/after the `assert False` line).
Instead, if I add an `assert False` in the caller (`groups`) the exception bubbles up.

mozci/task.py
```diff
@@ -262,33 +262,54 @@ def _load_errorsummary(self):
...
+
+        has_group_result = any(line["action"] == "group_result" for line in lines)
+
         for line in lines:
             if line["action"] == "test_groups":
                 groups = list(set(line["groups"]) - {"default"})
 
-            elif line["action"] == "test_result":
+            # After enough time has passed since https://bugzilla.mozilla.org/show_bug.cgi?id=1632242,
+            # we can switch to look for `group_result` lines exclusively.
```
ahal, line 276: Rather than linking to the bug, just add a `TODO` with a date, e.g:
```
# TODO After August 1st 2020, switch to using group_result exclusively.
```

mozci/task.py
```diff
@@ -262,33 +262,54 @@ def _load_errorsummary(self):
...
+                    group_results[group] = "ERROR"
+
+            elif line["action"] == "group_result":
+                group_results[line["group"]] = line["status"]
 
             elif line["action"] == "log":
                 self._errors.append(line["message"])
 
+        self._results = [
+            GroupResult(group, True if result == "OK" else False)
```
ahal, line 297: nit: `GroupResult(group, result == "OK")`

mozci/task.py
```diff
@@ -262,33 +262,54 @@ def _load_errorsummary(self):
...
+        ]
+
         if groups is not None:
             # Assume all groups for which we have no results passed.
-            # In practice, they could also have been skipped (we should ignore them when
-            # it will be feasible, https://github.com/mozilla/mozci/issues/166).
-            known_results = {result.group for result in self._results}
+            # After enough time has passed since https://bugzilla.mozilla.org/show_bug.cgi?id=1632242
+            # and https://bugzilla.mozilla.org/show_bug.cgi?id=1631515, we
+            # can stop assuming since all groups will have an associated result.
```
ahal, line 306: Ditto, just specify how much time is "enough" directly in the comment.

mozci/task.py
marco-c: Eheh, leftover from a previous iteration.

mozci/task.py
marco-c: Good idea, I've done that also for another case in the same file (I haven't done it here in particular because one of the two bugs isn't fixed yet).

Issue #189, opened 2020-05-12
Use adr.util.memoized_property instead of functools.lru_cache to avoid memory leaks


marco-c:
`functools.lru_cache` when used on a method will leak memory (when an instance is no longer reachable, both the instance and the cache will not be garbage collected).

marco-c:
We can possibly also stop using adr.util.memoized_property and just use functools.cached_property everywhere.

marco-c:
We can't use `functools.cached_property` yet as we don't want to limit the library to Python 3.8+ for now, so we'll have to use `adr.util.memoized_property`.

PR #193, opened 2020-05-13
Some memoization improvements

marco-c:
The main change is the fix for #189.


marco-c:
Note: the third commit here depends on https://github.com/mozilla/adr/pull/120.

marco-c:
> Note: the third commit here depends on mozilla/adr#120.

OK, green now, after updating adr.

PR #197, opened 2020-05-25
Don't consider classifications for failing tasks that contain a group, if the group didn't fail

marco-c:
When analyzing a given group, we were considering classifications belonging to failing tasks which contained the group whether the group failed or not. If the group didn't fail, we can be sure that the classification was not related to it.


marco-c:
I've added the test, I think it could be valuable in avoiding regressing this.

Issue #198, opened 2020-05-25
Some wrongly classified regressions can be missed


marco-c:
E.g. `ccd3a06efe81123e9ff32a697bc3c0f54e26a734` regressed `widget/tests/browser/browser.ini`, but:
- sheriffs marked the failing test-verify tasks on the push itself as "intermittent";
- in the following 6 pushes, the test passed;
- in the 7th push, it failed and was marked as "fixed by commit 332e06a545d2" (`332e06a545d2` is the backout of `ccd3a06efe81123e9ff32a697bc3c0f54e26a734`).

Since the initial classification is "intermittent", we consider the test as passing.
Even if we ignored "intermittent" classifications (or if we decided not to consider "intermittent" classifications when there is another classification for the same task/group that is "fixed by commit"), we'd still consider the test as passing as it passes in the following pushes.
If we both ignored "intermittent" classifications and analyzed further pushes instead of stopping at the first push with a pass, we'd catch the regression, but we'd also get two wrong results: `browser/base/content/test/fullscreen/browser.ini` and `browser/base/content/test/plugins/browser.ini`.

We'd get `browser/base/content/test/fullscreen/browser.ini` because there is a failure in a following push with a wrong "fixed by commit 332e06a545d2" classification (there is also a later failure of the same group which is marked as "intermittent").

We'd get `browser/base/content/test/plugins/browser.ini` because there is a failure in a following push with a wrong "fixed by commit 332e06a545d2" classification.

I propose we do the following:
1) look at all statuses between the push itself and its backout (#137);
2) if a task/group is marked as "intermittent" and then as "fixed by commit", consider it as a regression (#137);
3) if a task/group is marked as "fixed by commit" and then as "intermittent", don't consider it as a regression (same if there are multiple "fixed by commit" classifications pointing to different commits);
4) if a task/group still fails on the backout of the push, don't consider it as a regression of the push (#104);
5) if the status is PASS on some pushes and FAIL on some other pushes, consider it as INTERMITTENT (still considered as a regression of the push): #160. Unless the FAIL is marked as "intermittent" or "fixed by commit" pointing to an unrelated backout.

For this particular push, this should mean we'd get correctly get `widget/tests/browser/browser.ini` and wrongly get `browser/base/content/test/plugins/browser.ini`, but at least we wouldn't get `browser/base/content/test/fullscreen/browser.ini`.

[...]

PR #199, opened 2020-05-25
Misc improvements to the regression finding algorithm

marco-c:
I've identified a few improvements that could be done to the regression finding algorithm by comparing some of the regressions found so far at the label-level vs the group-level.

See #198.


marco-c:
> Though I have to admit that this is starting to get complex enough [...]

Agreed... I'm thinking a refactoring would be needed (maybe after fixing #204) to simplify things. We have a large enough number of tests that it should be relatively safe.

PR #201, opened 2020-05-25
Don't trust test-verify results, unless there is at least one group failure reported

marco-c:
This is based on top of #197.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -459,10 +459,16 @@ class Push:
             for group in task.groups:
                 config_groups[(task.configuration, group)].append(task)
 
-        return {
-            config_group: GroupSummary(config_group[1], tasks)
-            for config_group, tasks in config_groups.items()
-        }
+        config_group_summaries = {}
+        for config_group, tasks in config_groups.items():
+            group_summary = GroupSummary(config_group[1], tasks)
+            # TODO: We need this check because GroupSummary might ignore some test-verify
+            # tasks. A few months after https://bugzilla.mozilla.org/show_bug.cgi?id=1640758
+            # is fixed, we can remove it.
+            if len(group_summary.tasks) > 0:
+                config_group_summaries[config_group] = group_summary
+
+        return config_group_summaries
 
     @memoized_property
     def group_summaries(self):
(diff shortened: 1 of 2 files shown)


mozci/task.py
```diff
@@ -368,6 +368,16 @@ class GroupSummary(RunnableSummary):
     def __post_init__(self):
         assert all(self.name in t.groups for t in self.tasks)
 
+        # Because of https://bugzilla.mozilla.org/show_bug.cgi?id=1640758, we can't trust
+        # test-verify when there are no reported failures.
+        # TODO: Drop this filtering a few months after https://bugzilla.mozilla.org/show_bug.cgi?id=1640758 is fixed.
```
ahal, line 373: nit: no need to put url in comment twice.

mozci/push.py
```diff
@@ -459,11 +459,17 @@ def config_group_summaries(self):
...
+        config_group_summaries = {
             config_group: GroupSummary(config_group[1], tasks)
             for config_group, tasks in config_groups.items()
         }
 
+        return {
+            config_group: summary
+            for config_group, summary in config_group_summaries.items()
+            if len(summary.tasks) > 0
+        }
```
ahal, line 471: Rather than iterate over config_groups twice, I think it's more efficient and readable to use a for loop and build `config_group_summaries`.

mozci/push.py
```diff
@@ -480,7 +486,15 @@ def group_summaries(self):
...
 
-        return {group: GroupSummary(group, tasks) for group, tasks in groups.items()}
+        group_summaries = {
+            group: GroupSummary(group, tasks) for group, tasks in groups.items()
+        }
+
+        return {
+            group: summary
+            for group, summary in group_summaries.items()
+            if len(summary.tasks) > 0
```
ahal, line 496: Same here.

mozci/task.py
marco-c: Fixed

mozci/push.py
marco-c: I usually prefer comprehensions, but I guess here the advantage is lost as you need two... I've applied the change.

PR #206, opened 2020-06-01
Support tasks with multiple errorsummaries

marco-c:
test-verify tasks might have multiple errorsummary.log files.


marco-c:
Marked as draft as I need to handle the case where we have the same group in the different errorsummary.log files.

marco-c:
It should now be ready (I suggest hiding whitespace changes while reviewing).

PR #220, opened 2020-06-17
Add two missing WPT-based suites to the list of suites

marco-c:


marco-c:
Landing without review as it's an obvious fix :)

Issue #102, opened 2020-03-25
Add a way to get durations of shadow scheduler tasks


ahal:
Currently the dashboard is measuring the # of tasks scheduled, but we should probably track the total duration of all tasks instead. Some tasks are an order of magnitude longer than others and our metrics should take this into account.

The problem is that we need the total time of the "shadow scheduler" tasks, and those don't all run on the push. So we can't just sum all the `task.duration`s. We'll need to query the average duration from ActiveData or something. Since these should be somewhat stable, we also likely don't want to re-query them *too* often (maybe cache results for a day or two)..

An argument can be made that this should live in the `ETL`, though I think this data (especially a cache of average runtimes) would be useful to other consumers as well.

marco-c:
bugbug could use this data to penalize long running tasks (e.g. if it finds a test with "medium" confidence to fail, it might decide not to run it because of the cost).

ahal:
I guess we'll also want the durations of manifests as well eventually. Though fine to start with tasks I think.

ahal:
I believe a good way of implementing this is to add `duration` (both total and average) to the `Label/GroupSummary` classes. Then we can implement a static `create` method on both that accepts a date range as input and uses ActiveData to find all labels/groups within the range and returns the `Summary` object containing them. So from a user's view, it might look something like:

```
summary = LabelSummary.create(from="today", to="today-week")
print(summary.duration)  # total duration of all tasks
print(summary.average_duration)
```

We could potentially support other ways of instantiating this other than a date range (i.e, a revset) if we wanted.

klahnakoski:
>  have a helper that grabs all task durations

Yes, mozci is slow because it asks for data one-push-at-a-time. It would be faster to pull all the information from all pushes, all labels (all groups) at once.  Not sure how to do that; maybe the "helper" can do it, and fill the caches.

marco-c:
Blocks #1460 too.

klahnakoski:
Some steps:

1. (mozci) Get `duration` of every task
2. (mozci) Get `duration` of manifests (`max(test.end.time) - min(test.start_time)` for all test in manifest); be sure to consider tests with no manifest
3. (mozci) Calculate `overhead`: all the time not covered by manifests in the `(task.start_time, task.end_time)` range
4. (mozci) raise warnings if the manifest times overlap
5. (bigquery view)  every manifest and median duration
6. (bigquery view) average task `overhead` by ?suite?
[...]

PR #229, opened 2020-07-03
#102 - Add ability to obtain list of summary objects

worldomonation:
Changes:
  - implement a `make_summary_objects` API method to generate list of either `GroupSummary` or `LabelSummary` objects given a time frame and branch.

Example of use:
```
>>> from mozci.push import make_summary_objects
>>> label_results = make_summary_objects(branch="mozilla-central", from_date="today-day", to_date="today", type="label")
>>> label_results[0].durations
[3122, 3793, 4256]

>>> group_results = make_summary_objects(branch="mozilla-central", from_date="today-day", to_date="today", type="group")
>>> group_results[0].durations
[5.819999694824219, 8.485000133514404, 2.94700026512146, 7.112000226974486, 6.976000070571899, 12.858999729156494, 3.4600000381469727, 5.395999908447266, 6.997999906539917, 9.111000061035156, 7.062999725341797, 6.709000110626221, 8.343999862670898, 2.567000150680542, 17.61899971961975, 3.740000009536743, 2.7290003299713135, 8.474000215530396, 3.885000467300415, 8.652999877929688, 9.98099970817566, 2.6100006103515625, 3.062000036239624, 8.592999935150146, 17.656999826431274, 9.125, 5.954000234603882, 6.4770002365112305, 9.861999988555908, 5.726999759674073, 12.985000371932983, 2.583000421524048, 9.740000009536741, 6.801000118255615, 2.7350003719329834, 6.051000595092773]
```


worldomonation:
@ahal Is this a good first step to implement changes requested in #102 ?

mozci/task.py
```diff
@@ -441,6 +442,8 @@ class LabelSummary(RunnableSummary):
 
     label: str
     tasks: List[Task]
+    from_date: str = ""
+    to_date: str = ""
```
ahal: Maybe these were meant to go in the other commit? They don't appear used here.

mozci/task.py
```diff
@@ -5,6 +5,7 @@
 from dataclasses import dataclass, field
 from enum import Enum
 from inspect import signature
+from statistics import median
```
ahal, line 8: Nice, I'll admit I didn't even know about this library.

mozci/task.py
worldomonation: I have removed them in the second commit.

mozci/task.py
```diff
@@ -402,6 +402,18 @@ def classifications(self):
             and any(result.group == self.name and not result.ok for result in t.results)
         ]
 
+    @property
+    def durations(self):
+        return [t.duration for t in self.tasks]
```
ahal, line 407: Yeah, this one isn't right since we need to sum the durations of the "group" that this `GroupSummary` object corresponds to, not the tasks. That's what we'll need the AD query for.

mozci/task.py
```diff
@@ -401,6 +404,37 @@ def classifications(self):
...
+    @property
+    def durations(self):
+        def _subtract(tup):
+            return tup[0] - tup[1]
+
+        durations = []
+
+        for task in self.tasks:
+            data = run_query(
+                "group_summary_by_task_id", Namespace(task_id=task.id, group=self.name)
```
marco-c, line 416: We should try to make a single AD query rather than one per task.

If we implement what @ahal suggested in https://github.com/mozilla/mozci/issues/102#issuecomment-604464603:
> To limit queries to ActiveData, we could additionally have a helper that grabs all task durations in a single query and returns a list of `LabelSummaries` (similar to `make_push_objects` in the push.py module.

then this is less important.

mozci/task.py
```diff
@@ -401,6 +404,37 @@ def classifications(self):
...
+            )["data"]
+
+            durations.append(
+                sum(
+                    map(
+                        _subtract,
```
marco-c, line 422: Nit: use a lambda

mozci/queries/group_summary_by_task_id.query
```diff
@@ -0,0 +1,10 @@
+from: unittest
+select:
+    - result.test
```
marco-c, line 3: This seems unused

marco-c:
This would implement 1 and 2 from https://github.com/mozilla/mozci/issues/102#issuecomment-653527907.

Then, we'll need:
- a way to calculate the overhead (@klahnakoski, should it be task duration - duration of all manifests in the task?);
- a way to create LabelSummary/GroupSummary for a set of tasks. For non-manifest-based suites, we'll use LabelSummary (one LabelSummary per label, containing all tasks with the label from the past week). For manifest-based suites, we'll use GroupSummary (one GroupSummary per group and per configuration, containing all tasks that run the group on the configuration from the past week).

worldomonation:
Thanks for the review!

>     * a way to calculate the overhead (@klahnakoski, should it be task duration - duration of all manifests in the task?);

I have a follow-on patch to implement a `TestTask().overhead` property which is fairly simple, so I can possibly push it as part of this patch.

>     * a way to create LabelSummary/GroupSummary for a set of tasks. For non-manifest-based suites, we'll use LabelSummary (one LabelSummary per label, containing all tasks with the label from the past week). For manifest-based suites, we'll use GroupSummary (one GroupSummary per group and per configuration, containing all tasks that run the group on the configuration from the past week).

This I am not certain how to implement quite yet, but I think I can figure it out reasonably quickly.

mozci/task.py
ahal: Agreed, even a query per `GroupSummary` might be too expensive, though we'll need one as a fallback even in the event we implement the idea above. So let's not worry about the idea above for now and tackle it in a follow up PR.

But yeah, for now we should aim for a single query per `GroupSummary` instance (and ensure it gets memoized so we never call it twice).

mozci/task.py
```diff
@@ -401,6 +404,37 @@ def classifications(self):
...
+                        zip(data["result.end_time"], data["result.start_time"]),
```
klahnakoski, line 423: This may be simpler

    durations.append(sum(e-s for e, s in zip(data["result.end_time"], data["result.start_time"]))

ActiveData will accept a `{"format":"list"}`, `{"format":"table"}` and `{"format":"cubel"}` the last one is default, which is a columnar format, and what you are getting.  Sometimes this "columnar format is not useful.  Use `list`, and then your code is easier:

    durations.append(sum(r["end_time"]-r['start_time"] for r in data))

https://github.com/mozilla/ActiveData/blob/dev/docs/jx_tutorial.md#format-clause

mozci/push.py
```diff
@@ -1030,3 +1030,72 @@ def make_push_objects(**kwargs):
             cur._child = pushes[i + 1]
 
     return pushes
+
+
+def make_summary_objects(**kwargs):
```
ahal, line 1035: I know you were copying `make_push_objects` here, but would you mind not using `**kwargs` and including a docstring describing the parameters? Same for the two below. I'd like to fix up the `make_push_objects` function to do this as well.

mozci/task.py
```diff
@@ -451,6 +465,18 @@ def classifications(self):
             (t.classification, t.classification_note) for t in self.tasks if t.failed
         ]
 
+    @property
+    def durations(self):
+        return [t.duration for t in self.tasks]
```
ahal, line 470: These should have corresponding abstract properties on the base `RunnableSummary` class.

mozci/push.py
```diff
@@ -1030,3 +1030,72 @@ def make_push_objects(**kwargs):
...
+    except KeyError:
+        return []
+
+    # Sort by either the label or the name attribute.
+    # sort_attr = 'label' if kwargs.get('type').lower() == 'label' else 'name')
+    summaries.sort(key=lambda x: (getattr(x, "label", "name")))
+    return summaries
+
+
+def make_label_summary_objects(pushes, **kwargs):
```
ahal, line 1053: Would it be possible to turn these two function into private methods, and then have `make_summary_objects` as the single public entry point for this? I think it might make for a cleaner API.

mozci/queries/group_durations.query
```diff
@@ -0,0 +1,13 @@
+from: unittest
+groupby:
+    - task.id
+    - result.group
+select:
+    - {name: result_duration, value: result.duration, aggregate: sum}
```
ahal, line 6: Maybe Kyle can answer this better, but this need `aggregate: sum`? Shouldn't there only be a single duration per result? I could very well be mistaken.

mozci/push.py
worldomonation: Sure, I will make the change in an upcoming patch; in that I will use positional/named arguments. 

Also, generally speaking, more docstrings - I forgot about them this time for some reason.

mozci/push.py
worldomonation: To clarify, making these into a private method attached to the respective classes?

mozci/queries/group_durations.query
klahnakoski: The query is correct: We are adding the durations of all the tests in a group

mozci/task.py
worldomonation: Added as part of latest push.

worldomonation:
First comment updated with change description and use case reflecting current patch.

mozci/push.py
```diff
@@ -1030,3 +1030,98 @@ def make_push_objects(**kwargs):
...
+    pushes = make_push_objects(from_date=from_date, to_date=to_date, branch=branch)
+
+    summaries = globals()[func](pushes)
+
+    # Sort by either the label (LabelSummary) or name (GroupSummary).
+    summaries.sort(key=lambda x: (getattr(x, "label", "name")))
+    return summaries
+
+
+def __make_label_summary_objects(pushes):
```
worldomonation, line 1070: @ahal I've decided to simply add `__` in front to denote private method and keep these operations in this file.

I kept the `make_summary_objects` functions here because:
  - they operate on push(es)
  - `tasks.py` in my mind is scoped narrower than `push.py`

With that said, if you think it is better suited in `tasks.py`, I'm open to moving the entire set of functions that I've added in this patch there. I see them living as top-level functions in the file.

ahal: Yeah, that's what I had in mind (though usually a single underscore is the standard, double underscores typically denote magic methods on classes)

ahal: I agree having it in `tasks.py` makes sense as well.

PR #236, opened 2020-07-07
Make Push.generate_all_shadow_scheduler_tasks not throw an exception when one or more shadow schedulers failed on the push

### LINKED ISSUE #144: Push.generate_all_shadow_scheduler_tasks() shouldn't fall over if any given shadow-scheduler fails
`generate_all_shadow_scheduler_tasks()` is a generator that relies on `get_shadow_scheduler_tasks()` which appears to throw exceptions:  If any one scheduler messed up, collecting the others fail.  I suggest coding this function/generator so it provides as much information as possible; returning `None` for missing data or any other problems.  

If you do code in this none-safe way:  Consider shunting all exceptions to the logging channel as "warnings"; where "warning" is defined as an error that has a (non-optimal) mitigation.

marco-c:
Instead, return a set of tasks when a shadow scheduler did not fail, and an exception when a shadow scheduler failed.

Fixes #144

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -5,6 +5,7 @@ import itertools
 import math
 from argparse import Namespace
 from collections import defaultdict
+from typing import Iterator, Set, Tuple, Union
 
 import adr
 from adr.errors import MissingDataError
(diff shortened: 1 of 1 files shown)


mozci/push.py
```diff
@@ -998,7 +1001,10 @@ def generate_all_shadow_scheduler_tasks(self):
             if "shadow-scheduler" in label
         ]
         for name in sorted(names):
-            yield name, self.get_shadow_scheduler_tasks(name)
+            try:
+                yield name, self.get_shadow_scheduler_tasks(name)
+            except Exception as e:
+                yield name, e
```
ahal, line 1007: Please update the docstring as well.

marco-c: Done.

PR #241, opened 2020-07-13
#138 - Implement the concept of "data sources" in mozci

ahal:
This is a WIP PR not yet ready for review.

Rough steps for this work will be:

1. Stop depending on ADR for extra things (config, cache, memoize, logger, etc)
2. Implement "data sources", that is an interface for querying various pieces of data.
3. Refactor `adr` as a "data source" (alternatively pull out the bits of adr needed and stop depending on it)
4. Refactor `artifacts` as a "data source"
5. Implement a treeherder data source (likely in a separate PR)


mozci/__init__.py
```diff
@@ -2,7 +2,9 @@
 from pathlib import Path
 
 from adr import sources
-from adr.cli import setup_logging
+
+from mozci.configuration import config  # noqa
```
marco-c, line 2: Nit: why the noqa?

mozci/configuration.py
```diff
@@ -0,0 +1,171 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: I imagine this file was mostly copied over from `adr`, so I won't review it.

mozci/util/cache_stores.py
```diff
@@ -0,0 +1,245 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: Similarly for this file

mozci/util/cache_stores.py
```diff
@@ -0,0 +1,245 @@
...
+    def get(self, key):
+        value = super(RenewingFileStore, self).get(key)
+        if value is None:
+            return None
+
+        self.put(key, value, self.retention)
+        return value
+
+
+def get_taskcluster_options():
```
marco-c, line 131: Nit: could probably move to the taskcluster util module

mozci/__init__.py
ahal: Flake8 complains about unused imports, but I wanted this to be importable from the root `mozci` module. I guess we could also just instantiate it here rather than in `configuration.py`.

mozci/data/base.py
```diff
@@ -69,3 +69,11 @@ def get(self, name: str, **context: Any) -> Dict[Any, Any]:
         # Validate output.
         contract.validate_out(result)
         return result
+
+
+def register_sources():
```
marco-c, line 74: Were you thinking of allowing external users to define their own datasource or would they always use the built-in ones?

mozci/data/contract.py
```diff
@@ -3,18 +3,51 @@
...
 class Contract:
     name: str
+    description: str
     validate_in: Schema
     validate_out: Schema
 
 
-_contracts: Tuple[Contract] = (
-    Contract(name="placeholder", validate_in=Schema({}), validate_out=Schema({}),),
+_contracts: Tuple[Contract, ...] = (
```
marco-c, line 17: The only drawback I see is when two datasources underlying DBs are mismatching.
E.g. on AD maybe we can get the push_tasks data with one single query and the push_test_groups with a separate query, on Treeherder maybe it's two queries for push_tasks data and one of the two would also get data for push_test_groups.
So, our contracts will always be optimized better for one datasource rather than another.

mozci/data/sources/activedata/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+        retries = defaultdict(int)
+
+        for item in result["data"]:
+            if "id" not in item:
+                logger.trace(f"Skipping {item} because of missing id.")
+                continue
+
+            task_id = item["id"]
+
+            # If a task is re-run, use the data from the last run.
```
marco-c, line 47: Maybe this logic should not be in the datasource implementation but in the push module

marco-c: So it doesn't have to be repeated in all datasources

mozci/data/sources/activedata/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+    def run_push_test_groups(self, **kwargs):
+        result = adr.query.run_query(
+            "push_test_groups_from_unittest", Namespace(**kwargs)
+        )
+        groups = defaultdict(dict)
+        for item in self.normalize(result):
+            if "result_group" not in item:
+                continue
+
+            groups[item["id"]][item["result_group"]] = bool(item["result_ok"])
```
marco-c, line 102: Note there are cases where one of "result_group" or "result_ok" is missing, we should only store a result if they're both there

mozci/push.py
```diff
@@ -237,77 +237,36 @@ def tasks(self):
...
         if not was_cached:
             # Gather information from the unittest table. We allow missing data for this table because
             # ActiveData only holds very recent data in it, but we have fallbacks on Taskcluster
             # artifacts.
             try:
-                add(run_query("push_tasks_results_from_unittest", args, cache=False))
+                groups = data.handler.get(
+                    "push_test_groups", branch=self.branch, rev=self.rev
+                )
+                for task in tasks:
```
marco-c, line 259: This could be undefined if push_tasks throws MissingDataError

marco-c:
I've only done a quick review, I trust @armenzg to do another pass and @camd to check if the datasource contracts are relatively easy to implement with Treeherder.

mozci/data/contract.py
```diff
@@ -0,0 +1,54 @@
...
+    validate_in: Schema
+    validate_out: Schema
+
+
+_contracts: Tuple[Contract, ...] = (
+    Contract(
+        name="push_tasks",
+        description="Data about the tasks that ran on a given push.",
+        validate_in=Schema({Required("branch"): str, Required("rev"): str,}),
+        validate_out=Schema(
```
camd, line 22: This looks like a straightforward contract to comply with.  Not a problem in Treeherder.

mozci/data/contract.py
```diff
@@ -0,0 +1,54 @@
...
+                    Optional("duration"): int,
+                }
+            ]
+        ),
+    ),
+    Contract(
+        name="push_test_groups",
+        description="Data about the test groups that ran on a given push.",
+        validate_in=Schema({Required("branch"): str, Required("rev"): str,}),
+        validate_out=Schema(
```
camd, line 41: This will likely involve Treeherder downloading the artifact.  Though @ahal had mentioned that the same will likely be true for some other datasources.  So perhaps we won't need to implement this on the treeherder side.

camd:
Does this look ready now?  Unfortunately we won't get that review from armen at this point.

ahal:
Yeah, thanks for the nudge. I'll address marco's comments and land it (sorry for the delay).

mozci/data/sources/activedata/__init__.py
ahal: Maybe, I think I'd like to leave it here for now though, as I'm not sure that all datasources will even have the concept of a re-run.

In general I think we should be fairly fluid with the contracts going forward (change them to suit the needs of our data sources).

mozci/data/contract.py
ahal: Yeah, I think we shouldn't hesitate to refactor contracts into smaller ones or change them around if we need to. E.g, we could break this up into pieces, but if using AD still query the parts that aren't needed right away so they're available later.

ahal:
Marco added a new AD query, rebasing to get that to use the new system.

ahal:
Fyi the cron task failure that showed up on master is unrelated to this PR (I verified the same failure exists without it).

PR #243, opened 2020-07-16
#102 - Add overhead value to TestTask

worldomonation:
This implements the TestTask.overhead property that returns a float value representing non-group time.

Example Usage:

```
>>> import mozci
>>> from mozci.push import Push
>>> push = Push('b32bf0e2fa08e29748da69ac64fbe41659ad372c', branch='autoland')
>>> task = push.tasks[3]
>>> task.overhead
447.7479989528656
```

Or..

```
>>> for label, overhead in [(task.label, task.overhead) for task in push.tasks if type(task) == TestTask]:
...     print(label, overhead)
...
test-macosx1014-64/debug-mochitest-browser-chrome-e10s-4 141.60400009155273
test-linux1804-64/debug-mochitest-browser-chrome-fis-e10s-7 117.6800000667572
test-windows7-32/debug-web-platform-tests-reftest-e10s-2 581.414999961853
test-macosx1014-64/debug-xpcshell-e10s-2 110.23099994659424
test-macosx1014-64-shippable/opt-mochitest-devtools-chrome-e10s-3 118.6470000743866
test-linux1804-64-asan/opt-mochitest-browser-chrome-e10s-13 376.37199997901917
test-windows7-32/debug-mochitest-browser-chrome-e10s-3 293.8619999885559
test-linux1804-64-asan/opt-mochitest-browser-chrome-e10s-13 124.0220000743866
```

### DIFF
--- a/mozci/queries/test_task_overhead.query
+++ b/mozci/queries/test_task_overhead.query
@@ -1,0 +1,11 @@
+from: unittest
+format: list
+groupby: task.id
+limit: 20000
+select:
+    - {value: action.start_time, name: task_min, aggregate: min}
+    - {value: action.end_time, name: task_max, aggregate: max}
+    - {value: result.start_time, name: group_min, aggregate: min}
+    - {value: result.end_time, name: group_max, aggregate: max}
+where:
+    - in: {task.id: {$eval: task_id}}
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -522,6 +522,18 @@ class LabelSummary(RunnableSummary):
     def median_duration(self):
         return median(self.durations)
 
+    @property
+    def overheads(self):
+        return [task.overhead for task in self.tasks]
+
+    @property
+    def total_overheads(self):
+        return sum(self.overheads)
+
+    @property
+    def median_overhead(self):
+        return median(self.overheads)
+
     @memoized_property
     def status(self):
         overall_status = None
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -377,8 +377,8 @@ class TestTask(Task):
         """
         data = run_query("test_task_overhead", Namespace(task_id=self.id))["data"].pop()
         # Sanity check to ensure group start/end times are within task start/end.
-        assert data["task_min"] < data["group_min"]
-        assert data["task_max"] > data["group_max"]
+        if data["task_min"] < data["group_min"] or data["task_max"] > data["group_max"]:
+            logger.warning(f"task f{self.id} has inconsistent group duration.")
 
         return (data["group_min"] - data["task_min"]) + (
             data["task_max"] - data["group_max"]
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -2,6 +2,7 @@
 import json
 import os
 from abc import ABC, abstractmethod
+from argparse import Namespace
 from dataclasses import dataclass, field
 from enum import Enum
 from inspect import signature
@@ -9,6 +10,7 @@ from statistics import median
 from typing import Dict, List, Optional
 
 import requests
+from adr.query import run_query
 from adr.util import memoized_property
 from loguru import logger
 from urllib3.response import HTTPResponse
@@ -357,6 +359,30 @@ class TestTask(Task):
         # Remove the chunk number.
         parts = config.split("-")
         return "-".join(parts[:-1] if parts[-1].isdigit() else parts)
+
+    @property
+    def overhead(self):
+        """Calculate the overhead of a task.
+
+        The methodology is simple: each task (action) has a start/end time.
+        Each group also has a start/end time. Take the earlist known group start
+        and latest known group end time, ensure the two falls somewhere in between
+        task start/end.
+
+        Then calculate the overhead by taking the difference between the start
+        and end times.
+
+        Returns:
+            float: difference between task start/end and group star/end times.
+        """
+        data = run_query("test_task_overhead", Namespace(task_id=self.id))["data"].pop()
+        # Sanity check to ensure group start/end times are within task start/end.
+        assert data["task_min"] < data["group_min"]
+        assert data["task_max"] > data["group_max"]
+
+        return (data["group_min"] - data["task_min"]) + (
+            data["task_max"] - data["group_max"]
+        )
 
 
 # Don't perform type checking because of https://github.com/python/mypy/issues/5374.
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -365,15 +365,15 @@ class TestTask(Task):
         """Calculate the overhead of a task.
 
         The methodology is simple: each task (action) has a start/end time.
-        Each group also has a start/end time. Take the earlist known group start
+        Each group also has a start/end time. Take the earliest known group start
         and latest known group end time, ensure the two falls somewhere in between
         task start/end.
 
-        Then calculate the overhead by taking the difference between the start
-        and end times.
+        This definition of overhead does not take into account inter-group overhead
+        eg. restarting browser, teardown, etc.
 
         Returns:
-            float: difference between task start/end and group star/end times.
+            float: difference between task start/end and group start/end times.
         """
         data = run_query("test_task_overhead", Namespace(task_id=self.id))["data"].pop()
         # Sanity check to ensure group start/end times are within task start/end.


mozci/queries/test_task_overhead.query
```diff
@@ -0,0 +1,11 @@
+from: unittest
+format: list
+groupby: task.id
+limit: 20000
```
marco-c, line 4: Do we need such a high limit? If so, we might want to use `destination: url` like https://github.com/mozilla/mozci/blob/737c4cc0810bd745d13c90c511d18afff6baee20/mozci/queries/push_revisions.query.

mozci/task.py
```diff
@@ -358,6 +360,30 @@ def configuration(self):
         parts = config.split("-")
         return "-".join(parts[:-1] if parts[-1].isdigit() else parts)
 
+    @property
+    def overhead(self):
+        """Calculate the overhead of a task.
+
+        The methodology is simple: each task (action) has a start/end time.
+        Each group also has a start/end time. Take the earlist known group start
```
marco-c, line 368:
```suggestion
        Each group also has a start/end time. Take the earliest known group start
```

mozci/task.py
```diff
@@ -358,6 +360,30 @@ def configuration(self):
...
+        and latest known group end time, ensure the two falls somewhere in between
+        task start/end.
+
+        Then calculate the overhead by taking the difference between the start
+        and end times.
+
+        Returns:
+            float: difference between task start/end and group star/end times.
```
marco-c, line 376:
```suggestion
            float: difference between task start/end and group start/end times.
```

mozci/task.py
```diff
@@ -358,6 +360,30 @@ def configuration(self):
         parts = config.split("-")
         return "-".join(parts[:-1] if parts[-1].isdigit() else parts)
 
+    @property
+    def overhead(self):
```
marco-c, line 364: We discussed this, but I don't recall what we decided.
Should we:
1) consider overhead as the setup/teardown of the task (which this function is calculating);
2) consider overhead as the setup/teardown of the task + the amount of time elapsed between groups;
3) consider the setup/teardown overhead and overhead between groups separately.

CC @ahal @klahnakoski

marco-c, line 364: This is defined on a specific task, but we will have to define `overheads`, `total_overhead`, `median_overhead` on LabelSummary to actually use the information (e.g. the median overhead for a given suite on a given platform/configuration).

For task-level scheduling, we can just consider the sum of the median durations of all scheduled tasks.
For manifest-level scheduling, for each task we'll have to sum the durations of the groups in the task and sum the overhead associated with the task (I think by having a median overhead by suite and platform/configuration, we might be precise enough; maybe even just platform/configuration?).

worldomonation: This is the definition of overhead I used after discussing with @klahnakoski on Matrix.

I originally used the definition of 2. However, this ran into a problem with particular task `JCAGLvN0S7-XbnjhSmocyg` and possibly `xpcshell` as a whole.

The problem being, in that particular task, test manifests are run in parallel. So, taking the definition of 2, I would have calculated:

`task_duration - sum([group_end - group_start for group in task])`

This produced a total group runtime of somewhere between 5000-7000, depending on how the query was structured.
However, the `task_duration` of that particular task was 1861, so taking the difference resulted in a large negative value as the overhead.

With the definition of 1, it avoids the problem altogether, but we lose insight to what the inter-group overhead might be.

I put this patch up after suggestion that we have something that we can work with then iterate on it. Also because my last day prior to parental leave is tomorrow.

mozci/task.py
worldomonation: Thanks for spotting them. Sorry about the typos.

mozci/queries/test_task_overhead.query
worldomonation: No we do not - this was just carried over from the working query. It should return only 1 row.

mozci/task.py
worldomonation: Right, I will implement them while I wait for review from ahal and ekyle.

klahnakoski: @marco-c   I would prefer to go with `overhead` defined as

> our best estimate of the setup/teardown of the task + the amount of time elapsed between groups

Which makes the current algorithm an underestimate, and we can improve it later by including the inter group run times.

mozci/task.py
```diff
@@ -358,6 +360,30 @@ def configuration(self):
...
+        This definition of overhead does not take into account inter-group overhead
+        eg. restarting browser, teardown, etc.
+
+        Returns:
+            float: difference between task start/end and group start/end times.
+        """
+        data = run_query("test_task_overhead", Namespace(task_id=self.id))["data"].pop()
+        # Sanity check to ensure group start/end times are within task start/end.
+        assert data["task_min"] < data["group_min"]
+        assert data["task_max"] > data["group_max"]
```
klahnakoski, line 381: Do not crash on a problem:  We should be emitting problems as "warnings" to the logs and also declaring the overhead a zero (our best estimate).

Depanding on your logging library, something like:

    if not (data["task_min"] <= data["group_min"] <= data["group_max"] <= data["task_max"]):
        Log.warning(
            "task {{task.id}} has inconsistent group duration: {{data}}",  
            task=task,
            data=data
        )
        return 0

mozci/task.py
ahal: Agree going with 1) for now is easiest. We can iterate on it later to solve things like parallel execution. May need to update harnesses to adjust log messages to solve it.

klahnakoski: @worldomonation I will require the group-level run times, but I plan to compute the aggregates after the data has been put into the database.  Specifically, there is a BigQuery database with some of the data generated by `mozci`; my plan is to use include what these functions generate and form queries from there.

mozci/task.py
ahal: Why shouldn't we crash on a problem?

I agree crashing on a problem is bad if you're writing an ETL, but mozci is not an ETL, it's a library. Some consumers of the library will want to ignore errors and keep going (like an ETL). Other consumers of the library will want to fail loudly so they can be notified immediately when things go awry. The nice thing about crashing is that it let's the consumers decide how to handle the situation.

Though assertions are trickier to catch than a proper `Exception`, so if we're going to do this validation it should be using something like `ValueError` and the fact `ValueError` can be raised should be well documented.

mozci/task.py
```diff
@@ -358,6 +360,30 @@ def configuration(self):
         parts = config.split("-")
         return "-".join(parts[:-1] if parts[-1].isdigit() else parts)
 
+    @property
```
ahal, line 363: This should be memoized

ahal (changes requested):

mozci/task.py
klahnakoski: @ahal Since you are the owner, I will leave the error mitigation patterns to you. In the meantime, my reasons are:

* **We have a reasonable mitigation** - The `assert` is catching logic bugs: raising exception when the data is incoherent. The definition of `overhead` is our "best estimate of overhead": In the event of incoherent data, a zero estimate is a legitimate return value.
* **Less work for callers** - For every raised exception, there is at least one `try/catch` needed to handle that error.  All callers must implement their own mitigation, possibly the same mitigation. I doubt anyone will be writing an exception handler for these asserts.
* **Callers have less context** - Callers are less likely to know a suitable mitigation:  It is likely the caller will let the error propagate up to terminate the program.
* **asserts are for debugging** - These specific asserts exist to crash in the event of bad data; to crash in the event of unanticipated data; which is useful only in debugging. My suggestion is to shunt the error to `stderr` in production, and consume that channel offline to guide further development.
* **this is a rare event** - I do not expect these asserts to be broken

Yes, letting an exception escape the callee allows the caller to decide mitigation, but I am not advocating suppression of all exceptions. In this case, we have a reasonable mitigation strategy that I doubt any caller will ever write a handler for; plus it remains visible to anyone monitoring `stderr`.

Of course, I can always be wrong.  I am sure I have promoted past `warning`s to `error`s so that a better strategy can be implemented by a caller: In which case I advocate a more detailed description so the caller has the context to make the best decision:

    if not (data["task_min"] <= data["group_min"] <= data["group_max"] <= data["task_max"]):
        Log.error(
            "task {{task.id}} has inconsistent group duration: {{data}}",  
            task=task,
            data=data
        )

worldomonation:
I haven't forgotten about this PR yet, but I can't promise when I can get to it - job search takes first priority. 

I can likely address @ahal's comments on the method name and memoization but I think it would be better to not refactor the ActiveData query at this point in time.

PR #250, opened 2020-08-05
Only skip tasks which are chunked in the taskgraph when looking for regressions in pushes where manifest-level scheduling is enabled

### LINKED ISSUE #245: On manifest-level pushes, consider regressions of labels which are not manifest-level
#244 is OK as a temporary workaround while we run the manifest-level experiment.
In the longer term, if we want to make manifest-level the default, we need to ensure we can still find label-level regressions for tasks that do not have manifests (e.g. cppunit, gtest, build, reftest until it's manifest-level, etc.).

marco-c:
When manifest-level scheduling is enabled, we will still have some suites that are not chunked in the taskgraph (e.g. GTest).
We still need to be able to train the ML model on those suites, so we can't fully skip pushes where manifest-level scheduling
is enabled.

Fixes #245

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -299,6 +299,12 @@ class Push:
         except MissingDataError:
             pass
 
+        # Gather task tags from the task table.
+        try:
+            add(run_query("push_tasks_tags_from_task", args))
+        except MissingDataError:
+            pass
+
         # Let's gather error/results from cache or AD/Taskcluster
         test_tasks_results = adr.config.cache.get(self.push_uuid, {})
         was_cached = len(test_tasks_results.keys()) != 0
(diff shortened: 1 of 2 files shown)


mozci/push.py
```diff
@@ -522,13 +528,13 @@ def label_summaries(self) -> Dict[str, LabelSummary]:
...
-        # We can't consider tasks from manifest-level pushes for finding label-level regressions
-        # because tasks with the same name on different pushes might contain totally different tests.
-        if self.is_manifest_level:
-            return {}
-
         labels = defaultdict(list)
         for task in self.tasks:
+            # We can't consider tasks that were chunked in the taskgraph for finding label-level regressions
+            # because tasks with the same name on different pushes might contain totally different tests.
+            if "tests_grouped" in task.tags and task.tags["tests_grouped"] == "1":
```
ahal, line 535: nit: `task.tags.get("tests_grouped") == "1"`

mozci/push.py
```diff
@@ -299,6 +299,12 @@ def add(result):
         except MissingDataError:
             pass
 
+        # Gather task tags from the task table.
+        try:
+            add(run_query("push_tasks_tags_from_task", args))
```
ahal, line 304: This will conflict with my data_sources PR, but I guess this is needed to enable manifest scheduling so should take priority.

marco-c: Sorry :'(
Hopefully it's reasonably easy to rebase your PR

PR #251, opened 2020-08-05
Allow commits to continue after black / isort make modifications

ahal:
This change will silently incorporate the modifications that black / isort make into the commit (rather than failing and then requiring that the commit be re-run).

The author of pre-commit refuses to implement this and advises against doing this, as it means that what you have in your working directory isn't what ends up in the commit. While I do sympathize with this to some degree, in our case I found that 100% of the time was running:
```
$ git commit -a --amend
# black re-wrote stuff; commit fails
# didn't bother checking the modifications
$ git commit -a --amend
# now it works
```

Now the first `git commit -a --amend` will pass.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,11 +6,13 @@ repos:
 -   repo: https://github.com/timothycrosley/isort
     rev: 4.3.21
     hooks:
-    -   id: isort
+    - id: isort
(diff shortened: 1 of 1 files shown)


ahal:
@marco-c if this change bothers you, I'm fine dropping it. Not like it's a big deal to run the commit a second time.

PR #252, opened 2020-08-05
Support the new shadow-scheduler artifact format

ahal:


tests/test_push.py
```diff
@@ -370,20 +370,20 @@ def test_generate_all_shadow_scheduler_tasks(responses):
     )
 
     id = count(2)
-    for s in shadow_schedulers:
+    for ss in shadow_schedulers:
         s_id = next(id)
         responses.add(
             responses.GET,
-            get_index_url(push.index + f".source.shadow-scheduler-{s[0]}"),
+            get_index_url(push.index + f".source.shadow-scheduler-{ss[0]}"),
```
marco-c, line 377: Nit: since you're here, you could make push.index part of the f-string rather than string concatenate

PR #258, opened 2020-08-21
Create a Treeherder Datasource

camd:
@ahal - Does this look about right?  Please let me know if I'm on the right track or off in the weeds...  :)

### DIFF
--- a/mozci/data/sources/treeherder/__init__.py
+++ b/mozci/data/sources/treeherder/__init__.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+
+from collections import defaultdict
+
+from loguru import logger
+
+from mozci.data.base import DataSource
+
+try:
+    from treeherder.model.models import Job
+except ImportError:
+    Job = None
+
+
+class TreeherderSource(DataSource):
+    """Uses ORM to query Treeherder."""
+
+    name = "treeherder"
+    supported_contracts = ("push_tasks",)
+
+    @classmethod
+    def normalize(cls, jobs):
+        retries = defaultdict(int)
+        items = []
+
+        for job in jobs:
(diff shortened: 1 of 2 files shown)


camd:
@ahal This is now working on my end.  I have to do the Treeherder part to integrate it.  But would you want to re-review it prior to merge?

mozci/data/sources/treeherder/__init__.py
```diff
@@ -0,0 +1,87 @@
+# -*- coding: utf-8 -*-
+
+from collections import defaultdict
+
+from loguru import logger
+
+from mozci.data.base import DataSource
+
+try:
+    from treeherder.model.models import Job
```
ahal, line 10: You should add this as an optional dependency, e.g:
```
$ poetry add --optional treeherder
```
This will update the pyproject.toml and poetry.lock files automatically (just commit whatever they are after that).

ahal (approved):

mozci/data/sources/treeherder/__init__.py
ahal: Actually, looks like this package might not be on PYPI. Will this import only work on your machine? Would be nice to have more general support.

Maybe if you don't want to add this to pypi, we could add a configuration option that takes a path to the module which gets added to `sys.path` here?

camd: Yeah, Treeherder is not on PYPI.  This will only work when `mozci` is installed on a Treeherder instance.  The import references `Job` which is a Django model.  I wasn't sure what the best way was to do a *conditional* import for this.

I figured this way, it would try to do the import, and just punt if it can't find it.  Is adding it to `sys.path` cleaner?

PR #265, opened 2020-09-01
Don't ignore test-verify passes in pushes which contain the fix for bug 1640758

### LINKED ISSUE #207: Stop ignoring some test-verify tasks
Since https://bugzilla.mozilla.org/show_bug.cgi?id=1640758 has been fixed.

We should keep doing:
https://github.com/mozilla/mozci/blob/13807ef2f8dfd3284d871c949717e401febda532/mozci/task.py#L371-L379
only for pushes before the first push containing the fix.

How do we detect it? Maybe by using the push ID?

marco-c:
Fixes #207

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -394,6 +394,21 @@ class Push:
         """
         return self.task_labels - self.scheduled_task_labels
 
+    # Because of https://bugzilla.mozilla.org/show_bug.cgi?id=1640758, we can't trust
+    # test-verify when there are no reported failures.
+    # TODO: Drop this function a few months after the bug mentioned above is fixed.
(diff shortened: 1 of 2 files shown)


mozci/push.py
```diff
@@ -394,6 +394,21 @@ def unscheduled_task_labels(self):
         """
         return self.task_labels - self.scheduled_task_labels
 
+    # Because of https://bugzilla.mozilla.org/show_bug.cgi?id=1640758, we can't trust
+    # test-verify when there are no reported failures.
+    # TODO: Drop this function a few months after the bug mentioned above is fixed.
```
ahal, line 399: nit: please move this comment to the docstring and remove the duplicate comments below

PR #269, opened 2020-09-02
Create a Task object from the task data only after the data is complete

### LINKED ISSUE #267: Failure to create GroupSummary
STR:
```
from mozci.push import Push
push = Push("c076a08595452957f6084a8491ce94251dad35e8", branch="try")
push.get_likely_regressions("group")
```

It hits this assertion: https://github.com/mozilla/mozci/blob/678185ea9d4a673c78b4a71e3abcb58fd6b80123/mozci/task.py#L404,

that means a GroupSummary was created with at least one task that didn't contain the group itself.
It should never happen given how `GroupSummary`es are built: https://github.com/mozilla/mozci/blob/678185ea9d4a673c78b4a71e3abcb58fd6b80123/mozci/push.py#L424-L449.

The error is due to the wpt workaround.
The `name` in GroupSummary is `testing/web-platform/tests/css/cssom`, the group name in the task is `/css/cssom`.
The group name in the task should have been `testing/web-platform/tests/css/cssom` too.

marco-c:
So the Task's __post_init__ method has a chance to run on a complete object

Fixes #267, a regression from fdae1b03a73a9cb47900dd2cd0bedc0076742a67

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -248,7 +248,6 @@ class Push:
         tasks = []
         try:
             tasks = data.handler.get("push_tasks", branch=self.branch, rev=self.rev)
-            tasks = [Task.create(**task) for task in tasks]
         except MissingDataError:
             pass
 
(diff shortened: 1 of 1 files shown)


ahal (approved):
Thanks you!

PR #273, opened 2020-09-08
Don't consider intermittent failures which were already intermittent in a close parent as regressions

### LINKED ISSUE #1: An intermittent should not be considered a regression if it was already intermittent in one of its parents
It's not perfectly clear, as maybe a new intermittent was introduced on the current push, but it might be better not to consider intermittently failing tasks that were already intermittently failing in one of the parents.

marco-c:
Fixes #1

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -19,7 +19,15 @@ from mozci.errors import (
     ParentPushNotFound,
     PushNotFound,
 )
-from mozci.task import GroupResult, GroupSummary, LabelSummary, Status, Task, TestTask
+from mozci.task import (
+    GroupResult,
+    GroupSummary,
+    LabelSummary,
+    RunnableSummary,
+    Status,
+    Task,
+    TestTask,
+)
 from mozci.util.hgmo import HGMO
 
 BASE_INDEX = "gecko.v2.{branch}.revision.{rev}"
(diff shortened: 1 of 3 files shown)


mozci/task.py
```diff
@@ -362,6 +362,12 @@ def configuration(self):
 # Don't perform type checking because of https://github.com/python/mypy/issues/5374.
 @dataclass  # type: ignore
 class RunnableSummary(ABC):
+    @property
+    def is_intermittent(self):
+        return self.status == Status.INTERMITTENT or any(
+            c == "intermittent" for c, n in self.classifications
```
ahal, line 368: There's also an `autoclassified intermittent` classification.. not sure how much it actually gets used (if ever), but might as well support it I think.

mozci/push.py
```diff
@@ -894,20 +902,40 @@ def get_regressions(self, runnable_type: str) -> Dict[str, int]:
...
                     if name in runnable_summaries:
+                        found_in_parent = True
                         summary = runnable_summaries[name]
-                        if summary.status != Status.PASS and all(
-                            c != "intermittent" for c, n in summary.classifications
-                        ):
+
+                        # If the failure is not intermittent...
+                        if not failure_summary.is_intermittent:
+                            # ...and it failed permanently in the first parent where it run, it is a
```
ahal, line 916: nit: s/run/ran

ahal (approved):

mozci/task.py
marco-c: Is it set automatically or is it still chosen by sheriffs?

marco-c: (I'm wondering if we can trust it as we can trust `intermittent`)

marco-c: I've been told sheriffs don't use it anyway, so I'll not add it. Perhaps we can remove it from Treeherder options? CC @camd

PR #274, opened 2020-09-08
Make the maximum number of parents or children to look for previous/next task runs configurable

### LINKED ISSUE #8: Figure out if we need to analyze more pushes now that SETA switched from running all tasks every 5 pushes to every 9 pushes
We have a maximum number of pushes to analyze in the past and future, maybe we should re-evaluate the number we are using since the full pushes will be more rare.

marco-c:
Given that we have recently moved the backstop to be once every 20 pushes (https://hg.mozilla.org/mozilla-central/rev/17a6cf8777b1), we can't keep looking only in the previous/next 14 pushes.

Fixes #8

31 is a relatively big number, but in the end it should be fast enough when things are cached.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -32,7 +32,7 @@ from mozci.util.hgmo import HGMO
 
 BASE_INDEX = "gecko.v2.{branch}.revision.{rev}"
 
-MAX_DEPTH = 14
+MAX_DEPTH = config.get("maxdepth", 20)
 """The maximum number of parents or children to look for previous/next task runs,
 when the task did not run on the currently considered push.
 """
--- a/tests/test_regressions.py
+++ b/tests/test_regressions.py
@@ -692,7 +692,7 @@ def test_child_failed_and_bustage_fixed(create_pushes):
     p[len(p) - 1].tasks = [Task.create(id="1", label="test-prova", result="success")]
     p[len(p) - 1].bugs = {123}
 
-    assert p[i].get_regressions("label") == {"test-prova": 6}
+    assert p[i].get_regressions("label") == {"test-prova": 10}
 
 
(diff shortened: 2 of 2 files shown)


marco-c:
There are some test failures, I'll fix them after merging #273 and rebasing this.

marco-c:
@ahal this is now ready for review.

mozci/push.py
```diff
@@ -32,7 +32,7 @@
 
 BASE_INDEX = "gecko.v2.{branch}.revision.{rev}"
 
-MAX_DEPTH = 14
+MAX_DEPTH = 31
```
ahal, line 35: Isn't this a bit much? Shouldn't we just do 20 for the backstop?

marco-c: As discussed on Matrix, there are still cases (e.g. DONTBUILD) where we could go more than 20 pushes without a full back-stop, so 31 is to be safe (though it could even be larger than that in practice, it should be pretty rare). Moreover, it is also useful to find pre-existing intermittents in parents.

We don't want things to be too slow though, so we decided to make this configurable, and, as a follow-up, to add an option to use the previous backstop too (checking parameters.yml from the decision tasks.

mozci/push.py
```diff
@@ -32,7 +32,7 @@
 
 BASE_INDEX = "gecko.v2.{branch}.revision.{rev}"
 
-MAX_DEPTH = 14
+MAX_DEPTH = config.get("maxdepth", 20)
```
ahal, line 35: Was going to say you should add this config option to the docs.. but we don't have configuration page in the docs.. oops. I'll file an issue.

ahal (approved):

PR #281, opened 2020-09-17
Adjust group summarization to work on windows, accept labels+manifest…

jmaher:
…, and increase performance

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -1132,10 +1132,13 @@ def __make_group_summary_objects(pushes):
     """
     # Obtain all the task.id values contained in the pushes. It will be used in
     # the `where` query against ActiveData.
-    task_ids = [task.id for push in pushes for task in push.tasks]
-
-    # Extract the results of the above query.
-    results = run_query("group_durations", Namespace(task_id=task_ids))["data"]
+    results = []
+    revs = [p.rev for p in pushes]
+    try:
+        result = run_query("group_durations", Namespace(push_ids=revs))["data"]
+        results.extend(result)
+    except MissingDataError:
+        pass
 
     # Sort by the result.group attribute.
     results = sorted(results, key=lambda x: x[1])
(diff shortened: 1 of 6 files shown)


mozci/push.py
```diff
@@ -1132,10 +1132,15 @@ def __make_group_summary_objects(pushes):
...
-    task_ids = [task.id for push in pushes for task in push.tasks]
-
-    # Extract the results of the above query.
-    results = run_query("group_durations", Namespace(task_id=task_ids))["data"]
+    # task_ids = [task.id for push in pushes for task in push.tasks]
+    results = []
+    revs = [p.rev for p in pushes]
+    for rev in revs:
+        try:
+            result = run_query("group_durations", Namespace(push_ids=rev))["data"]
```
marco-c, line 1140: Instead of making one query per rev, could we do a single query with all the revs?

mozci/push.py
```diff
@@ -1146,11 +1151,16 @@ def __make_group_summary_objects(pushes):
...
     for task_id, result_group, result_duration in results:
         # Obtain TestTask object that matches the task_id.
-        task = [t for push in pushes for t in push.tasks if t.id == task_id].pop()
+        tasks = [t for push in pushes for t in push.tasks if t.id == task_id]
+        if len(tasks) == 0:
+            continue
+        task = tasks.pop()
 
+        parts = task.label.split('/')
+        key = "%s-%s:%s" % (parts[0], parts[1].split('-')[0], result_group)
```
marco-c, line 1160: If you want to get the task configuration, you can just use `task.configuration` instead

mozci/push.py
```diff
@@ -1146,11 +1151,16 @@ def __make_group_summary_objects(pushes):
...
         # Build the mapping of group to the group durations and TestTask objects.
-        mapping[result_group]["tasks"].append(task)
-        mapping[result_group]["durations"].append(result_duration)
+        mapping[key]["tasks"].append(task)
+        mapping[key]["durations"].append(result_duration)
 
     return [
         GroupSummary(key, value["tasks"], value["durations"])
```
marco-c, line 1166: `key` here is no longer a group now, I'm not sure this will work

mozci/push.py
```diff
@@ -1132,10 +1132,15 @@ def __make_group_summary_objects(pushes):
     """
     # Obtain all the task.id values contained in the pushes. It will be used in
     # the `where` query against ActiveData.
-    task_ids = [task.id for push in pushes for task in push.tasks]
-
-    # Extract the results of the above query.
-    results = run_query("group_durations", Namespace(task_id=task_ids))["data"]
+    # task_ids = [task.id for push in pushes for task in push.tasks]
```
marco-c, line 1135: Nit: you can remove this line now

mozci/push.py
marco-c: Ah, I see you changed the meaning of a GroupSummary.
Why though? We could keep the previous meaning, and have a GroupSummary for each group on a set of tasks.
E.g., to simplify, a GroupSummary for a given group for all Windows task, a GroupSummary for a given group for all Linux tasks, and so on.

jmaher:
I know @worldomonation was interested in seeing what changes I was making.

jmaher:
doing this in https://github.com/mozilla/mozci/pull/286 now

PR #286, opened 2020-09-23
Adjust group summary to be revision based, worked on windows, and acc…

jmaher:
…ount for configuration.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -1091,15 +1091,14 @@ def make_summary_objects(from_date, to_date, branch, type):
     # Obtain list of all pushes for the specified branch, from_date and to_date.
     pushes = make_push_objects(from_date=from_date, to_date=to_date, branch=branch)
 
-    summaries = globals()[func](pushes)
+    summaries = globals()[func](pushes, branch)
 
     # Sort by either the label (LabelSummary) or name (GroupSummary).
-    for summary in summaries:
-        summaries[summary].sort(key=lambda x: (getattr(x, "label", "name")))
+    summaries.sort(key=lambda x: (getattr(x, "label", "name")))
     return summaries
 
 
-def __make_label_summary_objects(pushes):
+def __make_label_summary_objects(pushes, branch):
     """Generates a list of LabelSummary objects from a list of pushes.
 
     Args:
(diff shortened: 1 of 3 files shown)


mozci/push.py
```diff
@@ -1112,13 +1113,21 @@ def __make_label_summary_objects(pushes):
         [task for push in pushes for task in push.tasks], key=lambda x: x.label
     )
 
-    mapping = defaultdict(lambda: defaultdict(list))
-
     # Make a mapping keyed by task.label containing list of Task objects.
+    configs = {}
```
marco-c, line 1117: Nit: `tasks_by_config` is clearer

mozci/push.py
```diff
@@ -1112,13 +1113,21 @@ def __make_label_summary_objects(pushes):
...
     for task in tasks:
-        mapping[task.label]["tasks"].append(task)
-
-    return [LabelSummary(key, value["tasks"]) for key, value in mapping.items()]
+        if task.configuration not in configs.keys():
+            # Dictionary to hold the mapping keyed by result.group mapped to list of
+            # task.id and list of and result.duration.
+            configs[task.configuration] = defaultdict(lambda: defaultdict(list))
+        configs[task.configuration][task.label]["tasks"].append(task)
```
marco-c, line 1123: Nit: I know it was there before too, but why not dropping "tasks" and just doing:
```
configs[task.configuration][task.label].append(task)
```
so `configs[task.configuration] = defaultdict(lambda: defaultdict(list))` can become `configs[task.configuration] = defaultdict(list)`. Or `configs` could become `defaultdict(lambda: defaultdict(list))` instead of `{}` and you could remove the `if task.configuration not in configs.keys():` check.

mozci/push.py
```diff
@@ -1112,13 +1113,21 @@ def __make_label_summary_objects(pushes):
...
+
+    retVal = {}
+    for c in configs:
+        retVal[c] = [
+            LabelSummary(key, value["tasks"]) for key, value in configs[c].items()
+        ]
+    return retVal
```
marco-c, line 1130: Nit:
```
return [LabelSummary(key, value["tasks"]) for config in configs for key, value in configs[config].items()]
```
or better if you do what suggested above:
```
return [LabelSummary(label, tasks) for config in tasks_by_config.keys() for label, tasks in tasks_by_config[config].items()]
```

mozci/push.py
```diff
@@ -1132,27 +1141,38 @@ def __make_group_summary_objects(pushes):
...
     # the `where` query against ActiveData.
-    task_ids = [task.id for push in pushes for task in push.tasks]
-
-    # Extract the results of the above query.
-    results = run_query("group_durations", Namespace(task_id=task_ids))["data"]
+    results = []
+    revs = [p.rev for p in pushes]
+    try:
+        result = run_query("group_durations", Namespace(push_ids=revs))["data"]
+        results.extend(result)
```
marco-c, line 1148: Nit:
```
results += run_query("group_durations", Namespace(push_ids=revs))["data"]
```
so it's more readable as you avoid the intermediate `result` variable

mozci/push.py
```diff
@@ -1132,27 +1141,38 @@ def __make_group_summary_objects(pushes):
...
-    # Dictionary to hold the mapping keyed by result.group mapped to list of
-    # task.id and list of and result.duration.
-    mapping = defaultdict(lambda: defaultdict(list))
-
+    configs = {}
     for task_id, result_group, result_duration in results:
         # Obtain TestTask object that matches the task_id.
-        task = [t for push in pushes for t in push.tasks if t.id == task_id].pop()
+        tasks = [t for push in pushes for t in push.tasks if t.id == task_id]
+        if len(tasks) == 0:
```
marco-c, line 1159: I'm surprised this can happen :S

mozci/push.py
```diff
@@ -1132,27 +1141,38 @@ def __make_group_summary_objects(pushes):
...
+        configs[task.configuration][result_group]["tasks"].append(task)
+        configs[task.configuration][result_group]["durations"].append(result_duration)
+
+    retVal = {}
+    for c in configs:
+        retVal[c] = [
+            GroupSummary(key, value["tasks"], value["durations"])
+            for key, value in configs[c].items()
+        ]
+    return retVal
```
marco-c, line 1178: Same nits as above :)

PR #296, opened 2020-09-30
Adjust group summaries to be revision based, support windows paths, and account for configuration instead of merging all data together

jmaher:


mozci/push.py
```diff
@@ -1132,27 +1131,38 @@ def __make_group_summary_objects(pushes):
     """
     # Obtain all the task.id values contained in the pushes. It will be used in
     # the `where` query against ActiveData.
-    task_ids = [task.id for push in pushes for task in push.tasks]
-
-    # Extract the results of the above query.
-    results = run_query("group_durations", Namespace(task_id=task_ids))["data"]
+    results = []
+    revs = [p.rev for p in pushes]
+    # if we have too many revsisions, activedata returns an error
```
marco-c, line 1136:
```suggestion
    # if we have too many revisions, ActiveData returns an error
```

mozci/push.py
```diff
@@ -1132,27 +1131,38 @@ def __make_group_summary_objects(pushes):
...
-    # Dictionary to hold the mapping keyed by result.group mapped to list of
-    # task.id and list of and result.duration.
-    mapping = defaultdict(lambda: defaultdict(list))
+    tasks_by_config = {}
+    task_id_to_task = {t.id: t for push in pushes for t in push.tasks}
 
     for task_id, result_group, result_duration in results:
-        # Obtain TestTask object that matches the task_id.
-        task = [t for push in pushes for t in push.tasks if t.id == task_id].pop()
+        # there are a few tasks where we do not find them, it is unclear why
```
marco-c, line 1151: Let's file an issue so we don't forget to investigate this

mozci/push.py
```diff
@@ -1112,16 +1112,15 @@ def __make_label_summary_objects(pushes):
...
-    mapping = defaultdict(lambda: defaultdict(list))
-
     # Make a mapping keyed by task.label containing list of Task objects.
+    tasks_by_config = defaultdict(lambda: defaultdict(list))
     for task in tasks:
-        mapping[task.label]["tasks"].append(task)
+        tasks_by_config[task.configuration][task.label]["tasks"].append(task)
 
-    return [LabelSummary(key, value["tasks"]) for key, value in mapping.items()]
+    return [LabelSummary(label, tasks) for config in tasks_by_config.keys() for label, tasks in tasks_by_config[config].items()]
```
marco-c, line 1120: Mmmh, I'm surprised Black is not acting here, the line looks too long

mozci/push.py
```diff
@@ -1132,27 +1131,38 @@ def __make_group_summary_objects(pushes):
...
+        tasks_by_config[task.configuration][result_group]["tasks"].append(task)
+        tasks_by_config[task.configuration][result_group]["durations"].append(result_duration)
 
     return [
         GroupSummary(key, value["tasks"], value["durations"])
-        for key, value in mapping.items()
-    ]
+        for config in tasks_by_config.keys()
+        for key, value in tasks_by_config[config].items()
+        ]
```
marco-c, line 1168: Similarly here, Black should complain as `]` is not indented as `return [`

mozci/push.py
marco-c: CC @ahal

marco-c (approved):
LGTM!
Can you write a nicer commit message (like the one you had a few PRs ago :P).

PR #304, opened 2020-10-05
Add a hg.mozilla.org-based data source to retrieve push revisions

marco-c:
Depends on #303.

PR #305, opened 2020-10-06
Add some additional methods to the taskcluster module

marco-c:
To support #304.

E.g. to get all tasks belonging to a given push:
```
decision_task_id = taskcluster.find_task_id(f"gecko.v2.{branch}.revision.{rev}.taskgraph.decision")
decision_task_data = taskcluster.get_task(decision_task_id)  # could avoid this request and always assume the decision_task_id is the same as the group ID.
tasks = get_tasks_in_group(decision_task_data["taskGroupId"])
```

### DIFF
--- a/mozci/util/taskcluster.py
+++ b/mozci/util/taskcluster.py
@@ -6,11 +6,13 @@
 import functools
 
 import taskcluster_urls as liburls
+from taskcluster.helper import TaskclusterConfig
 
 from mozci.util import yaml
 from mozci.util.req import get_session
 
 PRODUCTION_TASKCLUSTER_ROOT_URL = "https://firefox-ci-tc.services.mozilla.com"
+taskcluster_config = TaskclusterConfig(PRODUCTION_TASKCLUSTER_ROOT_URL)
 
 
 def _do_request(url, force_get=False, **kwargs):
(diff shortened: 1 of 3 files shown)


mozci/util/taskcluster.py
```diff
@@ -73,3 +75,26 @@ def get_index_url(index_path):
...
+def get_tasks_in_group(group_id):
+    tasks = []
+
+    def _save_tasks(response):
+        tasks.extend(response["tasks"])
+
+    queue = taskcluster_config.get_service("queue")
+    queue.listTaskGroup(group_id, paginationHandler=_save_tasks)
+
+    return tasks
```
marco-c, line 100: I could have implemented this using requests, but this is way easier.

mozci/util/taskcluster.py
```diff
@@ -6,11 +6,13 @@
 import functools
 
 import taskcluster_urls as liburls
+from taskcluster.helper import TaskclusterConfig
```
ahal, line 9: You'll need to `poetry add` the taskcluster module.

mozci/util/taskcluster.py
ahal: It's not that hard :p 
https://searchfox.org/mozilla-central/source/taskcluster/taskgraph/util/taskcluster.py#327

But agree that using the API seems nicer. Should we convert everything to use the API?

ahal (changes requested):

mozci/util/taskcluster.py
marco-c: > It's not that hard :p

Yeah, but it is cleaner (i.e. less code) this way :P

> But agree that using the API seems nicer. Should we convert everything to use the API?

Yeah, I'll file it as a good first issue.

mozci/util/taskcluster.py
marco-c: Oops, I guess I was lucky as `adr` depends on `taskcluster`

mozci/util/taskcluster.py
marco-c: I filed https://github.com/mozilla/mozci/issues/306.

ahal (approved):

PR #316, opened 2020-10-15
Add a Taskcluster-based data source to retrieve list of tasks which ran on a given push

marco-c:
@camd unfortunately I can't properly test the changes in the treeherder data source you wrote.

### DIFF
--- a/mozci/data/sources/treeherder/__init__.py
+++ b/mozci/data/sources/treeherder/__init__.py
@@ -81,8 +81,8 @@ class TreeherderSource(DataSource):
             tasks = {}
 
             for task in self.normalize(jobs):
-                if task.get("tags"):
-                    task["tags"] = {t["name"]: t["value"] for t in task["tags"]}
+                # TODO: Add support for tags.
+                task["tags"] = {}
 
                 tasks[task["id"]] = task
             return tasks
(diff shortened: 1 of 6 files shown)


ahal (approved):
This looks good to me. I'll leave it up to Cam to decide whether this needs testing for the treeherder aspects or not.

We should get some unit and integration tests up for these data sources (though the latter will be hard for the treeherder ORM, will likely need to mock it). But I'm happy to leave that for a follow-up commit.

mozci/data/sources/treeherder/__init__.py
```diff
@@ -78,21 +78,47 @@ def run_push_tasks(self, **kwargs):
...
         return {}
+
+    def run_push_tasks(self, branch, rev):
+        tasks = self._get_tasks(branch, rev)
+
+        return [
+            {
+                "id": task_data["id"],
+                "label": task_data["label"],
+                "tags": task_data["tags"],
```
camd, line 103: The only error I found so far is this line should be:

    "tags": task_data.get("tags", {}),

Because line 84 above shows `tags` may not exist in the dict.

camd (approved):
Works well once I got my cache repopulated (I had deleted it).  Just that one bug I mentioned.

PR #321, opened 2020-10-21
Improve error handling in data handler / sources

ahal:
This adds some custom exceptions for various problems in the data
handler. It will also gracefully try to use the next source in line if
ContractNotFilled was raised.


marco-c (approved):
We should also mozcify `MissingDataError` at some point

PR #325, opened 2020-10-23
Add a treeherder_client data source to retrieve classifications using the public API

### LINKED ISSUE #315: Get classifications from treeherder data source
This will depend on the API Cam added going live in production.

ahal:
Fixes #315

### DIFF
--- a/mozci/data/sources/activedata/__init__.py
+++ b/mozci/data/sources/activedata/__init__.py
@@ -20,7 +20,7 @@ class ActiveDataSource(DataSource):
     name = "adr"
     supported_contracts = (
         "push_tasks",
-        "push_tasks_results",
+        "push_tasks_classifications",
         "push_test_groups",
         "push_revisions",
     )
@@ -58,6 +58,17 @@ class ActiveDataSource(DataSource):
                 # We don't need to store the retry ID.
                 del item["retry_id"]
 
+            # Normalize result
+            result_map = {
+                "success": "passed",
+                "testfailed": "failed",
+                "bustage": "failed",
+                "usercancel": "canceled",
+                "retry": "exception",
+            }
+            if item.get("result") in result_map:
+                item["result"] = result_map[item["result"]]
+
             items.append(item)
         return items
 
@@ -73,6 +84,7 @@ class ActiveDataSource(DataSource):
             "id",
             "label",
             "result",
+            "state",
         )
 
         for task in self.normalize(result):
@@ -99,6 +111,8 @@ class ActiveDataSource(DataSource):
 
             tasks[task["id"]] = task
 
+        # TODO: Figure out why we have some results from the `push_tasks_tags_from_task` query
+        # that we don't have from `push_tasks_from_treeherder`.
         result = adr.query.run_query(
             "push_tasks_tags_from_task", Namespace(branch=branch, rev=rev)
         )
@@ -110,8 +124,6 @@ class ActiveDataSource(DataSource):
                 if any(k not in tag for k in ("name", "value")):
                     continue
 
-                # TODO: Figure out why we have some results from the `push_tasks_tags_from_task` query
-                # that we don't have from `push_tasks_from_treeherder`.
                 if item["id"] in tasks:
                     tasks[item["id"]]["tags"][tag["name"]] = tag["value"]
 
@@ -124,20 +136,21 @@ class ActiveDataSource(DataSource):
             {
                 "id": task_data["id"],
                 "label": task_data["label"],
+                "result": task_data["result"],
+                "state": task_data["state"],
+                "duration": task_data["duration"],
                 "tags": task_data["tags"],
             }
             for task_id, task_data in tasks.items()
         ]
 
-    def run_push_tasks_results(self, branch, rev):
+    def run_push_tasks_classifications(self, branch, rev):
         tasks = self._get_tasks(branch, rev)
 
         result = {}
         for task_id, task_data in tasks.items():
             result[task_id] = {
-                "result": task_data["result"],
                 "classification": task_data["classification"],
-                "duration": task_data["duration"],
             }
 
             if task_data.get("classification_note"):
--- a/mozci/data/sources/treeherder/__init__.py
+++ b/mozci/data/sources/treeherder/__init__.py
@@ -3,10 +3,12 @@
 from collections import defaultdict
 from functools import lru_cache
 
+from adr.util.memoize import memoized_property
 from loguru import logger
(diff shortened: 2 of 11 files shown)


ahal:
Not flagging for review just yet as I haven't done much testing on this (though it appears to work on my one test case I was using during development).

ahal:
@camd this PR changes the name of your data source from `treeherder` to `treeherder_db`, so you'd need to adjust your config file after it lands. I can make it backwards compatible if it makes things easier for you (though if it's not a big deal would prefer to change the name now while you're likely the only consumer of it).

camd:
No need for backward compat.  I'll just update my stuff.  I just have the one in-progress PR anyway.  Thanks man.

camd (approved):
Looks good to me.  👍

mozci/task.py
```diff
@@ -125,7 +125,8 @@ class Task:
     label: Optional[str] = field(default=None)
     duration: Optional[int] = field(default=None)
     result: Optional[str] = field(default=None)
-    classification: Optional[str] = field(default=None)
+    state: Optional[str] = field(default=None)
```
marco-c, line 128: What do we use the state for?

mozci/data/sources/treeherder/__init__.py
```diff
@@ -3,24 +3,71 @@
...
+        data = self._run_query(f"/project/{branch}/note/push_notes/?revision={rev}")
+
+        classifications = {}
+        for item in data:
+            job = item["job"]
+            classifications[job["task_id"]] = {
+                "classification": item["failure_classification_name"]
+            }
+            if item["text"]:
+                classifications[job["task_id"]]["failure_classification_name"] = item[
```
marco-c, line 58: This should be `classification_note`, right?
How come voluptous is not catching the schema mismatch?

mozci/task.py
ahal: Nothing at the moment. I just figured it would be useful, e.g if someone were inspecting a task and noticed there was no result, they might think why? Then notice that the state says "running".

I can also imagine applications where we'd want to filter out running tasks, and saying `task.state == "completed"` is nicer than `task.result is not None`

mozci/data/sources/treeherder/__init__.py
ahal: Good catch. Turns out the push I randomly selected to test on happened to not have any classification notes on it. I updated the test to make sure we validate the output against the contract.

mozci/data/sources/activedata/__init__.py
```diff
@@ -20,7 +20,7 @@ class ActiveDataSource(DataSource):
     name = "adr"
     supported_contracts = (
         "push_tasks",
-        "push_tasks_results",
+        # "push_tasks_classifications",
```
marco-c, line 23: I guess this was commented out for testing purposes?

mozci/data/sources/activedata/__init__.py
```diff
@@ -58,6 +58,17 @@ def normalize(cls, result):
...
                 del item["retry_id"]
 
+            # Normalize result
+            result_map = {
+                "success": "passed",
+                "testfailed": "failed",
+                "bustage": "failed",
+                "usercancel": "canceled",
+                "retry": "exception",
+            }
```
marco-c, line 68: `superseded` is not supported by AD?

mozci/data/sources/treeherder/__init__.py
```diff
@@ -3,24 +3,69 @@
...
+
+    @memoized_property
+    def session(self):
+        session = get_session()
+        session.headers = {"User-Agent": "mozci"}
+        return session
+
+    def _run_query(self, query):
+        query = query.strip("/")
+        url = f"{self.base_url}/{query}"
```
marco-c, line 38: Nit: you could use urljoin

mozci/data/sources/treeherder/__init__.py
```diff
@@ -3,24 +3,69 @@
...
+        if "?" in query:
+            url += "&format=json"
+        else:
+            url += "/?format=json"
```
marco-c, line 42: Nit: you could pass a params dict to `_run_query` (and then pass it down to requests.get), so you don't have to do this check but you can instead do `params["format"] = "json"`

mozci/data/sources/treeherder/__init__.py
```diff
@@ -54,7 +99,18 @@ def normalize(cls, jobs):
...
             }
-            items.append(result)
+
+            result_map = {
+                "success": "passed",
+                "testfailed": "failed",
+                "bustage": "failed",
+                "usercancel": "canceled",
+                "retry": "exception",
+            }
```
marco-c, line 109: treeherder too doesn't support superseded, maybe we can ignore superseded tasks altogether in the taskcluster data source too?

marco-c (approved):

mozci/data/sources/activedata/__init__.py
ahal: It is, it's just named the same so doesn't need translation.

mozci/data/sources/treeherder/__init__.py
ahal: Same as above, it does it just doesn't need translation.

ahal: Might as well keep it I guess... maybe we might want to get data on how often superseding happens?

mozci/data/sources/activedata/__init__.py
ahal: Yep, thanks for catching!

mozci/data/sources/treeherder/__init__.py
ahal: Good idea.

mozci/data/sources/treeherder/__init__.py
ahal: Hm, `urljoin` seems to assume that there's a base component, so e.g:

```
urljoin("foo/bar", "baz") => "foo/baz"
urljoin("foo/", "baz") => "foo//baz"
```

Think I'll just leave this one as is.

ahal:
Hm, that failed due to some kind of infra issue during `poetry install`... It passes for me locally, so think it's safe to merge anyway.

marco-c:
> Hm, that failed due to some kind of infra issue during poetry install... It passes for me locally, so think it's safe to merge anyway.

You can rerun the Travis build in such cases (it'll automatically update the status on GitHub too if the rerun ends up passing).

PR #332, opened 2020-10-28
Use json-automationrelevance to retrieve all information about commits

### LINKED ISSUE #331: Retrieving json-rev from hg.mozilla.org became super slow
json-rev now contains the full diff too, which is way more than we need.

marco-c:
Instead of relying on both json-rev and json-automationrelevance.

Fixes #331

This depends on adding `backedoutby` information in json-automationrelevance.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -64,7 +64,7 @@ class Push:
         if len(revs[0]) == 40:
             self.rev = revs[0]
         else:
-            self.rev = self._hgmo["node"]
+            self.rev = self._hgmo.node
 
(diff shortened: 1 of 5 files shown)


ahal (approved):
Lgtm! Does the backedoutby change on the server apply to older pushes as well and that's why we don't need to handle the case where it doesn't exist?

marco-c:
> Lgtm! Does the backedoutby change on the server apply to older pushes as well and that's why we don't need to handle the case where it doesn't exist?

Yep!

PR #335, opened 2020-10-29
Add a method to generate all groups, on given configurations, selected by shadow schedulers

### LINKED ISSUE #334: Add a method to gather all manifests selected by a shadow scheduler
We have `generate_all_shadow_scheduler_tasks`, but we need an equivalent for manifests.

marco-c:
Fixes #334

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -22,6 +22,7 @@ from mozci.task import (
     Status,
     Task,
     TestTask,
+    get_configuration_from_label,
 )
 from mozci.util.hgmo import HGMO
 
@@ -975,7 +976,7 @@ class Push:
         )
 
     @memoize
-    def get_shadow_scheduler_tasks(self, name: str) -> Set[str]:
+    def get_shadow_scheduler_tasks(self, name: str) -> List[dict]:
         """Returns all tasks the given shadow scheduler would have scheduled,
         or None if the given scheduler didn't run.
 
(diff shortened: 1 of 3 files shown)


ahal:
Travis failures look legit (likely just need to update the tests)

mozci/task.py
```diff
@@ -62,6 +62,18 @@ class Status(Enum):
 )
 
 
+def get_configuration_from_label(label: str) -> str:
```
ahal, line 65: Maybe you can refer to bug 1632870 here so if we end up fixing that and stumble across this comment, we can create an issue to switch over.

mozci/push.py
```diff
@@ -1002,14 +1011,35 @@ def generate_all_shadow_scheduler_tasks(
...
+        for name in sorted(self.shadow_scheduler_names):
             try:
-                yield name, self.get_shadow_scheduler_tasks(name)
+                yield name, set(
+                    t["label"] for t in self.get_shadow_scheduler_tasks(name)
+                )
+            except Exception as e:
+                yield name, e
+
+    def generate_all_shadow_scheduler_config_groups(
```
ahal, line 1022: Won't force you to block on it, but would be nice to have a test similar to the task version:
https://github.com/mozilla/mozci/blob/master/tests/test_push.py#L361

ahal (approved):
Looks good, please fix the test failures first (and consider adding a new test).

mozci/push.py
marco-c: Done!

mozci/task.py
marco-c: Done!

PR #336, opened 2020-10-29
Treeherder datasource push test groups

camd:
This adds `push_test_groups` contract support to the Treeherder Datasource.  It also fixes a couple things that were broken with the recent changes.  When a task is in-progress, it has a result of `unknown`.  And we get the result of `busted` rather than `bustage` in our Tasks.

### DIFF
--- a/mozci/data/sources/treeherder/__init__.py
+++ b/mozci/data/sources/treeherder/__init__.py
@@ -11,9 +11,11 @@ from mozci.errors import ContractNotFilled
 from mozci.util.req import get_session
 
 try:
-    from treeherder.model.models import Job
+    from treeherder.log_parser.failureline import get_group_results
+    from treeherder.model.models import Job, Push
 except ImportError:
     Job = None
+    Push = None
 
 
 class BaseTreeherderSource(DataSource):
@@ -65,6 +67,7 @@ class TreeherderDBSource(BaseTreeherderSource):
     supported_contracts = (
         "push_tasks",
         "push_tasks_classifications",
+        "push_test_groups",
     )
 
     @classmethod
@@ -181,3 +184,12 @@ class TreeherderDBSource(BaseTreeherderSource):
                 ]
 
         return result
+
+    def run_push_test_groups(self, branch, rev):
+        if not Push:
+            raise ContractNotFilled(
+                self.name, "push_test_groups", "could not import Push model"
+            )
+
+        push = Push.objects.get(repository__name=branch, revision=rev)
+        return get_group_results(push)
--- a/mozci/data/sources/treeherder/__init__.py
+++ b/mozci/data/sources/treeherder/__init__.py
@@ -144,6 +144,7 @@ class TreeherderDBSource(BaseTreeherderSource):
         return tasks
 
     def run_push_tasks(self, branch, rev):
+        keys = ("id", "label", "result", "duration", "tags", "state")
         if not Job:
             raise ContractNotFilled(
                 self.name, "push_tasks", "could not import Job model"
@@ -153,11 +154,9 @@ class TreeherderDBSource(BaseTreeherderSource):
 
         return [
             {
-                "id": task_data["id"],
-                "label": task_data["label"],
-                "result": task_data["result"],
-                "duration": task_data["duration"],
-                "tags": task_data["tags"],
+                key: task_data[key]
+                for key in keys
+                if not (key == "result" and task_data[key] == "unknown")
             }
             for task_id, task_data in tasks.items()
         ]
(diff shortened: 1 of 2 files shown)


mozci/data/contract.py
```diff
@@ -48,6 +48,7 @@ class Contract:
                         "exception",
                         "canceled",
                         "superseded",
+                        "unknown",
```
marco-c, line 51: In other data sources, we return a task object without a `result` if the task hasn't started yet or is still running, we should do the same for the Treeherder data source

mozci/data/sources/activedata/__init__.py
```diff
@@ -68,6 +68,8 @@ def normalize(cls, result):
             }
             if item.get("result") in result_map:
                 item["result"] = result_map[item["result"]]
+            else:
+                logger.error(item)
```
marco-c, line 72: I guess this one was a debugging statement?

mozci/data/sources/treeherder/__init__.py
```diff
@@ -103,6 +107,7 @@ def normalize(cls, jobs):
                 "success": "passed",
                 "testfailed": "failed",
                 "bustage": "failed",
```
marco-c, line 109: Do we also need to keep `bustage` in this map or is `busted` enough?

marco-c (commented):
Nit: It'd be nice to split the PR into four commits:
- one to add the "state" field;
- one to avoid setting "result" when the task is unscheduled or still running;
- one to add the implementation of the `push_test_groups` contract;
- one to improve the logging in the __post_init__ function of the TestTask class.

mozci/data/sources/treeherder/__init__.py
camd: `bustage` never comes into Treeherder.  Not sure if that's needed for something else though.

> Nit: It'd be nice to split the PR into four commits:
> 
>     * one to add the "state" field;
> 
>     * one to avoid setting "result" when the task is unscheduled or still running;
> 
>     * one to add the implementation of the `push_test_groups` contract;
> 
>     * one to improve the logging in the **post_init** function of the TestTask class.

Sure, I can split it to those commits.  👍

mozci/data/sources/activedata/__init__.py
camd: Yes, removed.

mozci/data/contract.py
camd: Done

mozci/data/sources/treeherder/__init__.py
camd: Done

mozci/data/sources/treeherder/__init__.py
```diff
@@ -152,11 +158,9 @@ def run_push_tasks(self, branch, rev):
...
         return [
             {
-                "id": task_data["id"],
-                "label": task_data["label"],
-                "result": task_data["result"],
-                "duration": task_data["duration"],
-                "tags": task_data["tags"],
+                key: task_data[key]
+                for key in keys
+                if not (key == "result" and task_data[key] == "unknown")
```
marco-c, line 163: This should cause an exception if there are classifications, as the returned objects would contain some keys that the contract does not expect

marco-c: Oh forget it, I see you have a list of keys and you are not doing task_data.keys ;)

mozci/data/sources/treeherder/__init__.py
```diff
@@ -182,3 +185,16 @@ def run_push_tasks_classifications(self, branch, rev):
...
+            push = Push.objects.get(repository__name=branch, revision=rev)
+            result = get_group_results(push)
+
+            return result
+
+        logger.trace(
+            "Unable to reach Treeherder as a datasource because the Push model "
+            "was not available to import."
+        )
+        return {}
```
marco-c, line 200: I'd do as we do for the other contract implementations in this class, `raise ContractNotFilled` if we can't import the model

marco-c (approved):

mozci/data/sources/treeherder/__init__.py
marco-c: I've checked in ActiveData too, the `bustage` string never occurs. Probably a typo by @ahal, so we can remove it altogether (I'll remove it from the ActiveData source too).

mozci/data/sources/treeherder/__init__.py
camd: Shoot, my bad.  OK, fixed this.

marco-c (approved):

PR #365, opened 2020-11-30
Don't memoize the Task artifacts property

marco-c:
It is only used once, there is no need to memoize it.


jmaher (approved):
simple and straightforward

PR #367, opened 2020-11-30
Don't load all errorsummary.log lines in memory

### LINKED ISSUE #366: OOMs while analyzing some pushes
E.g. f6f29c9cc744bff2c94903bed8b625c8cbac9118.

I think this is due to the fact that errorsummary.log files became way bigger since last week (probably because of some bug).

marco-c:
Fixes #366

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -2,4 +2,4 @@
 multi_line_output=3
 include_trailing_comma=True
 line_length=88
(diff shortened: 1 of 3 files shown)


jmaher (approved):
thanks for the comment to remove old_style_group_results, this looks good.  a little bit of rubber stamping here.  Ideally I will know this code better in the coming weeks.

PR #369, opened 2020-12-02
Cache all push tasks

### LINKED ISSUE #368: Cache results from data sources
Previously, when relying on AD, we were caching all query results.
Now, when using other data sources, we are not caching results from them.

marco-c:
Fixes #368

Previously, we were relying on adr caching the results of AD queries, so the `data.handler.get` calls in the `tasks` function were relatively quick as they were hitting the cache.
Now, this is not happening anymore with the different data sources that we have introduced. So if we run a script which accesses `push.tasks` twice, the second will not be fast as it used to be.

Also now that the `data.handler.get` calls are not cached, we don't need to worry about storing the same data twice, and so we can cache the whole tasks list instad of only caching the errorsummary.log results ,

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -248,6 +248,10 @@ class Push:
         Returns:
             list: A list of `Task` objects.
         """
+        tasks = config.cache.get(f"{self.push_uuid}/tasks")
+        if tasks is not None:
+            return tasks
+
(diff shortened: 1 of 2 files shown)


jmaher (approved):
this change looks straightforward.

Issue #390, opened 2021-01-05
Reinstate integration tests


marco-c:
They are not running anymore, they are currently depending on ActiveData.
We should:
1) Remove the dependency on ActiveData, just use our other data sources;
2) Make them run daily.

These tests are kind of important to ensure there are no regressions in mozilla-central that could taint our datasets.

ahal:
This PR will temporarily remove the busted tests, but we can look at history to get them back:
https://github.com/mozilla/mozci/pull/433

Issue #359, opened 2020-11-19
Stop depending on ActiveData for group durations


marco-c:
https://github.com/mozilla/mozci/blob/216e9f036e4ee4d18276d78c9e153309b96ce409/mozci/push.py#L1154-L1161

marco-c:
Since https://bugzilla.mozilla.org/show_bug.cgi?id=1671422, we have group duration information in the errorsummary.log files.
E.g. https://firefoxci.taskcluster-artifacts.net/NbUm06JOTbuwl5fqW3UkGQ/0/public/test_info//mochitest-plain_errorsummary.log:
```
{"action": "test_groups", "line": 2, "groups": ["dom/console/tests/mochitest.ini", "dom/performance/tests/mochitest.ini", "dom/events/test/mochitest.ini", "testing/mochitest/tests/Harness_sanity/mochitest.ini"]}
{"status": "OK", "duration": 6457, "line": 3422, "group": "dom/console/tests/mochitest.ini", "action": "group_result"}
{"status": "OK", "duration": 269498, "line": 3422, "group": "dom/performance/tests/mochitest.ini", "action": "group_result"}
{"status": "OK", "duration": 222151, "line": 3422, "group": "dom/events/test/mochitest.ini", "action": "group_result"}
{"status": "OK", "duration": 311901, "line": 3422, "group": "testing/mochitest/tests/Harness_sanity/mochitest.ini", "action": "group_result"}
```

marco-c:
Depends on #646.

PR #433, opened 2021-04-06
Remove dependency on adr

ahal:


mozci/push.py
```diff
@@ -1037,123 +1036,3 @@ def make_push_objects(**kwargs):
...
-    revs = [p.rev for p in pushes]
-    # if we have too many revisions, ActiveData returns an error
-    for i in range(0, len(revs), 30):
-        for revs_chunk in revs[i : i + 30]:
-            try:
-                results += adr.query.run_query(
-                    "group_durations", Namespace(push_ids=revs_chunk, branch=branch)
-                )["data"]
-            except adr.MissingDataError:
-                pass
```
ahal, line 1126: Oh, I just noticed this has a fallback. @marco-c  is this still being used?

tests/test_integration.py
```diff
@@ -65,118 +57,6 @@ def test_create_pushes_and_get_regressions():
     push.get_likely_regressions("group")
 
 
-def test_missing_manifests():
```
marco-c, line 68: Do we already have an issue on file to reimplement these without AD?

mozci/push.py
marco-c: The plan was to use this to show durations instead of # of groups for shadow scheduler evaluation, but we never ended up using it.
https://github.com/mozilla/mozci/issues/359 is tracking rewriting this without AD, but for the time being we can safely remove it.

mozci/push.py
```diff
@@ -1037,123 +1036,3 @@ def make_push_objects(**kwargs):
...
-    pushes = make_push_objects(from_date=from_date, to_date=to_date, branch=branch)
-
-    summaries = globals()[func](pushes, branch)
-
-    # Sort by either the label (LabelSummary) or name (GroupSummary).
-    summaries.sort(key=lambda x: (getattr(x, "label", "name")))
-    return summaries
-
-
-def __make_label_summary_objects(pushes, branch):
```
marco-c, line 1077: This one isn't using AD, so we might keep it

tests/test_integration.py
marco-c: We can use https://github.com/mozilla/mozci/issues/390 to track this.

marco-c:
Closing in favor of #663.

PR #436, opened 2021-04-07
Refactor TestTask._load_errorsummary into a data source

ahal:


sarah-clements:
@ahal I did some debugging/walking through the mozci code for retrieving likely regressions and it's making quite a few calls to hg. First to retrieve the parent push, followed by a query of that parent revision to retrieve the push head. Probably the first two are unavoidable since Treeherder doesn't store the parent push (I'm not even sure if it's part of the pulse payload).

Where I'm seeing there could be some optimizations is on line 845 in `push.get_regressions`. For each candidate regression, json pushes are being fetched individually whereas it looks like they can be retrieved by a range of ids instead. For ex: `https://hg.mozilla.org/mozilla-central/json-pushes?version=2&startID=38350&endID=38355` outside of the for loop, rather than `n` calls to hg for each slice of that id range.

I tried profiling that block of the code and got some undecipherable results... seems a lot of time is spent in the schema validators??? I might try a different debugger tool tomorrow - spent a good chunk of the day trying to understand the codebase and stepping through code.

ahal:
Thanks for the debugging Sarah! I'll file a new issue for looking into that.

mozci/data/sources/treeherder/__init__.py
```diff
@@ -192,7 +212,7 @@ def run_push_tasks_classifications(self, branch, rev):
 
         return result
 
-    def run_push_test_groups(self, branch, rev):
+    def get_push_test_groups(self, branch, rev) -> Dict[str, List[str]]:
```
marco-c, line 215: This is not implementing the new contract, right?

marco-c: Ah you moved it to a base class, forget it.

mozci/data/sources/treeherder/__init__.py
```diff
@@ -18,15 +21,31 @@
     Push = None
 
 
-class BaseTreeherderSource(DataSource):
-    pass
+class BaseTreeherderSource(DataSource, ABC):
+    lock = threading.Lock()
+    groups_cache: Dict[str, List[str]] = {}
```
marco-c, line 26: The only drawback I see here is that this will leak memory. We might want to delete the tasks from the cache once we get the groups from them (IIRC we only get them once anyway)

mozci/data/sources/artifact/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+from collections import defaultdict
+from typing import Any, Dict
+
+import requests
+
+from mozci.data.base import DataSource
+from mozci.util.taskcluster import get_artifact, list_artifacts
+
+
+class ErrorSummarySource(DataSource):
```
marco-c, line 13: We could potentially merge this into the Taskcluster datasource, and provide a way to prioritize contracts and data sources idependently, butI think it's not worth the additional complexitiy.

mozci/task.py
```diff
@@ -273,108 +279,25 @@ def __post_init__(self):
...
-        assert not is_no_groups_suite(self.label)
-
-        try:
-            paths = [a for a in self.artifacts if a.endswith("errorsummary.log")]
-        except IndexError:
-            return
-
-        groups = set()
-        group_results = {}
-        # TODO After April 1st 2021, switch to using group_result exclusively.
```
marco-c, line 296: I'm glad we can finally remove this!

mozci/task.py
```diff
@@ -273,108 +279,25 @@ def __post_init__(self):
...
-        # since the artifact is already downloaded, parsed and we need to
-        # iterate over it anyway. It doesn't really hurt and simplifies some
-        # logic. It also ensures we don't attempt to load the errorsummary more
-        # than once.
-        self._results = []
-        self._errors = []
-
-        # Make sure that we don't try to load errorsummary.log for suites which
-        # don't support groups.
-        assert not is_no_groups_suite(self.label)
```
marco-c, line 287: I think it'd be good to keep this assertion, maybe move it in the `groups` property?

mozci/data/sources/artifact/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+            if line["action"] == "test_groups":
+                groups |= set(line["groups"]) - {"default"}
+
+            elif line["action"] == "group_result":
+
+                group = line["group"]
+                if group not in group_results or line["status"] != "OK":
+                    group_results[group] = line["status"]
+
+            elif line["action"] == "log":
```
marco-c, line 60: We are not using errors yet anywhere as far as I remember, maybe it would be best to simply get rid of them and reduce memory usage (in some errorsummary.log, there can be GBs of logs)? True, they might be useful at some point (and that's why we added them in the first place), but for the time being... Anyway, I'll leave the decision to you.

mozci/task.py
```diff
@@ -273,108 +279,25 @@ def __post_init__(self):
...
-        # TODO After April 1st 2021, we can remove this assumption altogether, as all errorsummary.log
-        # files will have 'group_result' entries.
-        if not has_group_result and not has_crashed and len(groups) > 0:
-            self._results += [
-                GroupResult(group, True)
-                for group in groups
-                if group not in group_results
-            ]
-
-        self.__post_init__()
```
marco-c, line 360: We needed to run `__post_init__` because it's running some checks and rewriting some groups. I guess you could move the call to the `results` property, the line after `self._results = [`.

mozci/data/sources/artifact/__init__.py
ahal: Makes sense.. though I think if we remove them I'd like it to be outside of this PR. That way fewer things are changing at once.

marco-c: :+1:

mozci/data/sources/treeherder/__init__.py
ahal: I added an LRU dict library to cache this (I added a dependency rather than roll our own since it's implemented in C and is faster)

mozci/data/sources/artifact/__init__.py
ahal: Wouldn't be a bad feature.. but yeah definitely out of scope for this PR

mozci/data/sources/artifact/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+from mozci.data.base import DataSource
+from mozci.util.taskcluster import get_artifact, list_artifacts
+
+
+class ErrorSummarySource(DataSource):
+    name = "errorsummary"
+    supported_contracts = ("test_task_groups", "test_task_errors")
+
+    TASK_CACHE: Dict[str, Any] = defaultdict(
```
marco-c, line 17: We need to use lru here too, right?

mozci/data/sources/treeherder/__init__.py
```diff
@@ -35,6 +59,7 @@ def session(self):
         session.headers = {"User-Agent": "mozci"}
         return session
 
+    @lru_cache(maxsize=50)
```
marco-c, line 62: Why the cache here?

marco-c (approved):

mozci/data/sources/treeherder/__init__.py
ahal: I realized that if Treeherder is missing task ids (which seems to be the case sometimes), then we'll re-run the query again for each task id that is missing. This is a guard to ensure we aren't actually hitting Treeherder each time.

ahal:
Latest push was just rebase of poetry.lock.

mozci/data/sources/artifact/__init__.py
ahal: Done in latest commit.

PR #444, opened 2021-04-16
Grab json-pushes for multiple pushes at a time when iterating over children/parents

### LINKED ISSUE #443: Batch calls to hgmo when detecting regressions
Sarah found out that we are spending a lot of time querying `hg.mozilla.org` when detecting regressions and pointed out that we can gather data for much data at a time:
https://hg.mozilla.org/mozilla-central/json-pushes?version=2&startID=38350&endID=38355

Since we know how many pushes we will look into the future or past (MAX_DEPTH), we should be able to grab all the `hg.m.o` data we need in a single query.

ahal:
Fixes #443

### DIFF
--- a/mozci/util/hgmo.py
+++ b/mozci/util/hgmo.py
@@ -1,17 +1,17 @@
 # -*- coding: utf-8 -*-
 from __future__ import annotations
 
-from typing import Dict, Tuple
+from typing import Any, Dict, Tuple
 
 import requests
 from lru import LRU
 
 from mozci.errors import PushNotFound
-from mozci.util.memoize import memoize, memoized_property
+from mozci.util.memoize import memoized_property
 from mozci.util.req import get_session
 
 
-class HGMO:
+class HgRev:
     # urls
     BASE_URL = "https://hg.mozilla.org/"
     AUTOMATION_RELEVANCE_TEMPLATE = (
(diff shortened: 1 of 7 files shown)


mozci/util/hgmo.py
```diff
@@ -25,7 +26,8 @@ class HGMO:
     )
 
     # instance cache
-    CACHE: Dict[Tuple[str, str], HGMO] = {}
+    CACHE: Dict[Tuple[str, str], HgRev] = LRU(1000)
+    JSON_PUSHES_CACHE: Dict[int, Dict[str, Any]] = LRU(1000)
```
marco-c, line 30: Nit: You could define a Push object like: `HgPush = NewType("HgPush", Dict[str, Any])`

mozci/util/hgmo.py
```diff
@@ -36,20 +38,66 @@ def __init__(self, rev, branch="autoland"):
...
-        HGMO.CACHE[key] = instance
+        if key in HgRev.CACHE:
+            return HgRev.CACHE[key]
+        instance = HgRev(rev, branch)
+        HgRev.CACHE[key] = instance
         return instance
 
-    def _get_resource(self, url):
+    @staticmethod
+    def _get_and_cache_pushes(branch, url):
```
marco-c, line 48: Nit: can you add type information to the new functions you added? It should be simple in these cases

marco-c (approved):

marco-c:
@ahal could you tag a new release after merging this?

mozci/util/hgmo.py
ahal: Oops, yes I need to get in the habit more.

PR #450, opened 2021-04-20
Remove circular dependency between push and task objects

### LINKED ISSUE #449: RecursionError: maximum recursion depth exceeded while pickling an object
```
from_date = datetime(2021, 4, 1)
to_date = datetime(2021, 4, 19)

pushes = tuple(
    make_push_objects(
        from_date=from_date.strftime("%Y-%m-%d"),
        to_date=to_date.strftime("%Y-%m-%d"),
        branch="autoland",
    )
)
for push in pushes:
    push.tasks
```
fails with:
```
Traceback (most recent call last):
  File "/home/marco/Documenti/workspace/mozci/runpush.py", line 26, in <module>
    push.tasks
  File "/home/marco/Documenti/workspace/mozci/mozci/util/memoize.py", line 49, in __get__
    setattr(instance, name, self.func(instance))
  File "/home/marco/Documenti/workspace/mozci/mozci/push.py", line 293, in tasks
    config.cache.put(
  File "/home/marco/.local/lib/python3.9/site-packages/cachy/repository.py", line 88, in put
    self._store.put(key, val, minutes)
  File "/home/marco/.local/lib/python3.9/site-packages/cachy/stores/file_store.py", line 97, in put
    value = encode(str(self._expiration(minutes))) + encode(self.serialize(value))
  File "/home/marco/.local/lib/python3.9/site-packages/cachy/contracts/store.py", line 121, in serialize
    return self._serializer.serialize(data)
  File "/home/marco/Documenti/workspace/mozci/mozci/util/cache_stores.py", line 259, in serialize
    return self.compressor.compress(pickle.dumps(data))
RecursionError: maximum recursion depth exceeded while pickling an object
```

This is very likely a regression from d35ba971c04d088ae452972ddf7670c5828fb94d.

marco-c:
Fixes #449, a regression from d35ba97

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -277,7 +277,7 @@ class Push:
 
         # Gather group data.
         logger.debug(f"Gathering test groups for {self.rev}...")
-        concurrent.futures.wait(
+        done, _ = concurrent.futures.wait(
             [
                 Push.THREAD_POOL_EXECUTOR.submit(
                     lambda task: task.retrieve_results(self), task
@@ -287,6 +287,8 @@ class Push:
             ],
             return_when=concurrent.futures.FIRST_EXCEPTION,
         )
+        for f in done:
+            f.result()
 
         logger.debug(f"Retrieved all tasks and groups which run on {self.rev}.")
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -236,7 +236,7 @@ class Push:
             Task: A `Task` instance representing the decision task.
         """
         index = self.index + ".taskgraph.decision"
-        return Task.create(push=self, index=index)
+        return Task.create(index=index)
 
     @memoized_property
     def tasks(self):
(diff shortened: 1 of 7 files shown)


mozci/push.py
```diff
@@ -273,13 +273,15 @@ def tasks(self):
...
-        tasks = [Task.create(push=self, **task) for task in tasks]
+        tasks = [Task.create(**task) for task in tasks]
 
         # Gather group data.
         logger.debug(f"Gathering test groups for {self.rev}...")
         concurrent.futures.wait(
             [
-                Push.THREAD_POOL_EXECUTOR.submit(lambda task: task.groups, task)
+                Push.THREAD_POOL_EXECUTOR.submit(
+                    lambda task: task.retrieve_results(self), task
```
ahal, line 283: I get why you're setting it up this way, and I guess we can land this as is for now to unblock things..

But I think I'd prefer storing `rev` and `branch` on the Task object since that will allow consumers to instantiate task objects directly. We could even create a new contract to grab `rev` / `branch` from taskcluster in the event that it wasn't filled in from the `create`. This would allow consumers to do stuff like:
```
task = Task(<id>)
print(task.results)
```

without requiring the context of a `Push` class to use. While we don't currently have consumers doing this, I'm often sent a taskcluster link and asked to investigate the given task id, so could imagine it being useful in that case. I could imagine other non-human consumers using this as well.

It can't increase the cache size *that* much can it? Are we currently pickling the `Push` class once per task and storing it in cache? If so storing just the `rev` / `branch` is already a big improvement :p

marco-c: We can definitely try and see how large it is before/after with a bunch of pushes.

With the `Push` reference, it's likely that we were storing the Push object only once for all tasks belonging to a push. IDK if Pickle will do the same for primitive values such as strings.

ahal (approved):

PR #455, opened 2021-04-30
Taskcluster push tasks fix

### LINKED ISSUE #454: Builds not retrieved from treeherder_db source in v2.0.1
In the 2.0.1 release, something seems to have changed with regards to the `treeherder_db` source that causes build tasks not to be retrieved. The results I see locally (set to ['treeherder_client', 'hgmo', 'taskcluster']) is not the same as what is used in production ['treeherder_db', 'hgmo', 'taskcluster']. I don't see this discrepancy with v2.0.0 (results are the same across primary data sources).

ahal:
Fixes #454

### DIFF
--- a/mozci/data/sources/taskcluster/__init__.py
+++ b/mozci/data/sources/taskcluster/__init__.py
@@ -64,7 +64,7 @@ class TaskclusterSource(DataSource):
                 run = result["status"]["runs"][-1]
 
                 # Normalize the state to match treeherder's values.
-                if task["state"] in ("failed", "exception"):
+                if task["state"] == "failed":
                     task["state"] = "completed"
 
                 # Derive a result from the reasonResolved.
@@ -77,7 +77,7 @@ class TaskclusterSource(DataSource):
                     task["result"] = "exception"
                 else:
                     # Task is not finished, so there is no result yet.
-                    assert task["state"] in ("pending", "running")
+                    assert task["state"] in ("pending", "running", "exception")
 
                 # Compute duration.
                 if "started" in run and "resolved" in run:
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -15,6 +15,7 @@ from mozci.errors import (
     MissingDataError,
     ParentPushNotFound,
     PushNotFound,
+    SourcesNotFound,
 )
 from mozci.task import (
     GroupSummary,
@@ -287,8 +288,19 @@ class Push:
             ],
             return_when=concurrent.futures.FIRST_EXCEPTION,
         )
+        failed = set()
         for f in done:
-            f.result()
+            try:
+                f.result()
+            except SourcesNotFound as e:
+                task = e.context["task"]
+                failed.add(f"{task.id} - {task.label}")
+
+        if failed:
+            failed_str = "  \n".join(sorted(failed))
+            logger.warning(
+                f"Failed to get test groups from the following tasks:\n  {failed_str}"
+            )
 
         logger.debug(f"Retrieved all tasks and groups which run on {self.rev}.")
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -285,6 +285,7 @@ class Push:
                 )
                 for task in tasks
                 if isinstance(task, TestTask)
+                if task.state == "completed"
             ],
             return_when=concurrent.futures.FIRST_EXCEPTION,
         )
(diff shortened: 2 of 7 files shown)


mozci/push.py
```diff
@@ -287,8 +288,19 @@ def tasks(self):
             ],
             return_when=concurrent.futures.FIRST_EXCEPTION,
         )
+        failed = set()
         for f in done:
-            f.result()
+            try:
+                f.result()
+            except SourcesNotFound as e:
```
marco-c, line 296: Instead of ignoring these errors, couldn't we avoid trying to retrieve groups for such tasks?

marco-c: I guess this is what your next commit is doing at https://github.com/mozilla/mozci/pull/455/commits/92dc36226900dd8115a0802e6ae9e2d3dd61b51a#diff-cbdcefec12b605b952245761dbfbbbb2ec7fc73addcc510c2b8dc48a60bd5d02R288? Why is this still necessary then?

mozci/data/sources/taskcluster/__init__.py
```diff
@@ -64,7 +64,7 @@ def run_push_tasks(self, branch, rev):
                 run = result["status"]["runs"][-1]
 
                 # Normalize the state to match treeherder's values.
-                if task["state"] in ("failed", "exception"):
+                if task["state"] == "failed":
```
marco-c, line 67: `exception` could also mean a task timed out. Are we 100% sure we won't have errorsummary.log files in those cases? I guess so, just want to double confirm.

mozci/push.py
ahal: There are still edge cases that hit this, even with the next patch. I can't remember off-hand what task it was.. I guess it was missing from the "no groups suite" list. Either way, I think we want this guard since we don't want to regress `mozci` every time someone adds a new suite that doesn't have groups. Especially when groups might not even be needed for the specific use case (like in Sarah's case).

mozci/data/sources/taskcluster/__init__.py
ahal: Yeah, timed out tasks don't have an errorsummary.log. Though this timed out task has stated `failed` rather than `exception`. Which makes it tricky as sometimes `failed` tasks will have errorsummaries and sometimes they won't. I guess the guard from the previous commit is necessary for this reason too then...

mozci/push.py
marco-c: The data source could just return an empty list when the task doesn't have groups

marco-c (approved):

mozci/data/sources/taskcluster/__init__.py
ahal: If you were reading timeout in the taskcluster docs, I think they mean a different timeout when the job fails to start before the claim expires.

mozci/push.py
ahal: In the treeherder source's case we purposefully [raise ContractNotFilled](https://github.com/mozilla/mozci/blob/master/mozci/data/sources/treeherder/__init__.py#L44) so that we fallback to the artifact data source on older pushes that are no longer in the treeherder DB.

We could make the artifact data source return an empty list, but then what if you don't have it configured in your `config.toml`? Am I missing a downside to catching and warning here?

mozci/data/sources/taskcluster/__init__.py
marco-c: Tasks can fail with "exception" also when the deadline is exceeded (I've seen it with bugbug data gathering tasks).

mozci/push.py
marco-c: The only downsides are the additional (even if limited) complexity of the logic and the possibility of ignoring some real errors (e.g. say treeherder suddenly stops returning results for all tasks, we won't notice until we look at the logs).
They're relatively limited downsides though, so feel free to land this as-is :)

ahal: I think the root issue with your suggestion is that there's no proper way of distinguishing between the "push isn't in the DB" and "the task has no groups" in the Treeherder API. I guess we could check if there exists some result that gets returned, but that method of distinguishing would fail when e.g there is a build only push.

mozci/data/sources/taskcluster/__init__.py
ahal: Interesting.. I wonder what the difference is between those tasks and the one I linked. Either way a timeout should mean no errorsummary as that gets copied to the artifact dir *after* the suite finishes running.

PR #517, opened 2021-10-09
Convert all functions in the taskcluster module to use the taskcluster library

### LINKED ISSUE #306: Convert all functions in the taskcluster module to use the taskcluster library
See https://github.com/mozilla/mozci/pull/305#discussion_r500186280.

chichi012:
Fix #306

### DIFF
--- a/mozci/util/taskcluster.py
+++ b/mozci/util/taskcluster.py
@@ -10,6 +10,7 @@ from mozci.util import yaml
 from mozci.util.req import get_session
 
 PRODUCTION_TASKCLUSTER_ROOT_URL = "https://firefox-ci-tc.services.mozilla.com"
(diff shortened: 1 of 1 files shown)


chichi012:
I couldn't get the `get_artifact()` function and `find_task_id()` to work with the taskcluster library as tests were failing.

For `get_artifact()`, i think the taskcluster library only returns a url to download the artifact so the `_do_request()` is needed to stream/download the file. (not sure about this though)

For `find_task_id()`  it looks like one of the tests in test_task.py, specifically the test `test_create(response)` mocks the  get_index_url() function call and expects that it is called in `find_task_id`. Looks like I will have to update this test to use taskcluster too for this to work.

I need guidance on how to proceed with this but I think we can submit this PR now as this changed some of other functions to use the taskcluster. I can work on the other changes in a future PR.

PR #518, opened 2021-10-10
Add ability to skip tasks based on tier

### LINKED ISSUE #496: Add ability to include tier 3 tasks
We currently filter out tier 3 tasks at the data source level:
https://github.com/mozilla/mozci/blob/master/mozci/data/sources/taskcluster/__init__.py#L43
https://github.com/mozilla/mozci/blob/master/mozci/data/sources/treeherder/__init__.py#L151

I think in most cases this is the desired outcome, but there may be instances where folks *want* to be able to inspect them. I propose a two step fix here:
1. Stop filtering them out in the data sources, and instead filter them out in the `Push` class
2. Create an instance variable on the `Push` class to control whether tier 3 tasks should count or not

chichi012:
Fixes #496 

- Added a `tier` variable to `configuration.py`, defaulting to 2
- Stopped filtering Tier 3 tasks from the data sources in `treeherder` and `taskcluster`
- Used instance variable `config.tier` on the Push class to control which tier tasks should be filtered out. 
- Fixed failing test in `create_task` class

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -2,4 +2,4 @@
 multi_line_output=3
 include_trailing_comma=True
 line_length=88
-known_third_party = adr,appdirs,boto3,botocore,cachy,loguru,lru,pytest,requests,responses,taskcluster,taskcluster_urls,tomlkit,validx,yaml,zstandard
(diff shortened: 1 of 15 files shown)


chichi012:
@ahal I wrote the test `test_push_tasks_with_tier`. However, I noticed that `tasks` list is always empty regardless of what we have in the mock response.

PR #526, opened 2021-10-22
Classify regressions and push using bugbug data

### LINKED ISSUE #504: Use `bugbug` classifications to help determine regression likelihood
We have a machine learning model that assigns relevant tests to a push. The theory is that if the model things a test is highly relevant to a push, then any failure in that test is much more likely to be a regression.

Conversely, if the ML thinks a test is irrelevant to a push then any failures in that test are much more likely to be an intermittent.

We can use this data to improve the `get_likely_regressions` function. We could possibly also provide a dedicated function that just uses this data if it would be useful.

To solve this we'll need to create a new data contract plus a `bugbug` data source that fulfills it. Here's an example of a client querying for this data from bugbug:
https://searchfox.org/mozilla-central/source/taskcluster/taskgraph/util/bugbug.py#75

Once we have the data integrating it into the regression detection algorithm is a bit less scoped out.

La0:
Closes #504 

Algorithm implemented in `classify_regressions` for a push  is :
1. get schedules from bugbug (or local cache)
2. get likely regressions from mozci  (or local cache)
3. sort bugbug groups into high & low confidence buckets (excluding the medium ones)
4. detect if likely regressions is a cross config failure or not
5. get all groups that fail
6. build the group lists:
  - real failures = `likely_regressions & groups_cross_config_failure & groups_high`
  - intermittent failures = `groups_non_cross_config_failure & groups_low`
  - unknown failures = remaining failures not in the two previous lists

A second method `classify` simply gets those list and extrapolates a single value for the whole push:
- `PushStatus.BAD` when there are real failures 
- `PushStatus.GOOD` when there are only intermittent failures
- fallback to `PushStatus.UNKNOWN`

:warning: the `test.py` at the root level should be removed before merge, it's just for debugging (and would be implemented cleanly in https://github.com/mozilla/mozci-tools/issues/15).

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -5,6 +5,7 @@ import itertools
 import math
 from argparse import Namespace
 from collections import defaultdict
+from enum import Enum
 from typing import Dict, Iterator, List, Optional, Set, Tuple, Union
 
 from loguru import logger
@@ -37,6 +38,12 @@ when the task did not run on the currently considered push.
 """
 
 FAILURE_CLASSES = ("not classified", "fixed by commit")
+
+
+class PushStatus(Enum):
+    GOOD = 0
+    BAD = 1
+    UNKNOWN = 2
 
 
 class Push:
@@ -960,7 +967,16 @@ class Push:
         )
 
     def classify_regressions(self):
-        cache_prefix = f"{self.push_uuid}/test_bastien/"
+        """
+        Use group classification data from bugbug to classify all likely
+        regressions into three categories: real, intermittent or unknown failures
+
+        Output: a dict with several lists of task groups:
+        - real: set of tasks that are definitely real failures
+        - intermittent: set of tasks that are definitely intermittents
+        - unknown: set of tasks we don't have enough information about
+        """
+        cache_prefix = f"{self.push_uuid}/classify_group_tasks/"
 
         # Preset confidence thresholds from MC
         # CT_LOW = 0.7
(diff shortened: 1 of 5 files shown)


tests/test_push.py
```diff
@@ -49,6 +49,26 @@
...
+                    result="failed",
+                    _results=[GroupResult(group=f"group{i}", ok=False)],
+                )
+                for j in range(1, 4)
+            ],
+        )
+        for i in range(1, 6)
+    ]
+}
+CONFIGS = [f"test-task{j}" for j in range(1, 4)]
```
marco-c, line 70: Nit: TASKS

mozci/push.py
```diff
@@ -959,6 +976,140 @@ def get_likely_regressions(self, runnable_type):
...
+            # Link all the failing configurations on the given groups
+            return {
+                name: self.group_summaries[name].failing_configurations
+                for name in groups
+            }
+
+        # Output real, intermittent and unknown groupfailures
+        # along with their failing configurations
+        return Regressions(
+            real=_map_failing_configurations(real_failures),
```
marco-c, line 1085: Actually, we don't want to return the configuration names, but the tasks (which is what we are going to use for the treeherder API later)

marco-c: Basically something like this is what we want to return in `real`, `intermittent` and `unknown`:
```Python
[
    task for task in self.tasks
    if any(not result.ok and result.group == self.name for result in task.results)
]
```

(where `self` is GroupSummary)

marco-c: Sounds good.

marco-c (approved):

ahal (approved):

PR #543, opened 2021-11-08
Support classifying a range of pushes in the classify command

### LINKED ISSUE #535: Add the possibility for the `classify` command to classify a range of pushes
It could be useful to compare the classify command results with the actual decisions from sheriffs and see how often classify is correct.

EvaBardou:
Closes #535

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -44,37 +44,43 @@ class ClassifyCommand(Command):
 
     def handle(self):
         branch = self.argument("branch")
+        from_date = self.option("from-date")
+        to_date = self.option("to-date")
 
-        if not self.option("rev"):
-            from_date = self.option("from-date")
-            to_date = self.option("to-date")
+        if not (bool(self.option("rev")) ^ bool(from_date or to_date)):
+            self.line(
+                "<error>You must either provide a single push revision with --rev or define --from-date AND --to-date options to classify a range of pushes.</error>"
+            )
+            return
+
+        if self.option("rev"):
+            pushes = [Push(self.option("rev"), branch)]
+        else:
             if not from_date or not to_date:
                 self.line(
-                    "<error>You must either provide a single push revision with --rev or define --from-date AND --to-date options to classify a range of pushes.</error>"
+                    "<error>You must provide --from-date AND --to-date options to classify a range of pushes.</error>"
                 )
                 return
-            else:
-                try:
-                    datetime.datetime.strptime(from_date, "%Y-%m-%d")
-                except ValueError:
-                    self.line(
-                        "<error>Provided --from-date should be a date in yyyy-mm-dd format.</error>"
-                    )
-                    return
 
-                try:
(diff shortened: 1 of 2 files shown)


mozci/console/commands/push.py
```diff
@@ -28,16 +30,52 @@ def handle(self):
...
         {--output= : Path towards a directory to save a JSON file containing classification and regressions details in.}
     """
 
     def handle(self):
+        branch = self.argument("branch")
+
+        if not self.option("rev"):
+            from_date = self.option("from-date")
+            to_date = self.option("to-date")
+            if not from_date or not to_date:
```
ahal, line 51: Thanks for the validation! I think we should put this check outside of the `if not self.option("rev"):` statement though. Otherwise if the user specifies both `--rev` and `--to-date` we'll never reach this validation.

mozci/console/commands/push.py
```diff
@@ -56,61 +94,66 @@ def handle(self):
...
-                    "classification": classification.name,
-                },
-                "failures": {
-                    "real": {
-                        group: [
-                            {"task_id": task.id, "label": task.label}
-                            for task in failing_tasks
-                        ]
-                        for group, failing_tasks in regressions.real.items()
+        for push in pushes:
```
ahal, line 97: I noticed `cleo` has a progress feature that could be useful here:
https://cleo.readthedocs.io/en/latest/helpers/progress_bar.html

Though I guess if there is a ton of output being emitted from `mozci` it won't work. We'd need to only show the progress bar if other output is suppressed.. probably not worth implementing.

ahal (changes requested):
Thanks, looks like a nice change! Just a minor comment to fix up.

mozci/console/commands/push.py
EvaBardou: From my own understanding, the user should be able to only provide either a revision OR both from and to dates. When the revision is explicitly given, we only want to classify it and nothing else. But it really depends and what you want to achieve and it should be easily refactorable if you want me to support alternative cases.

mozci/console/commands/push.py
La0: I think that would be a nice followup, but it mayneed to be integrated a bit more deeply into mozci

mozci/console/commands/push.py
ahal: Yeah, that's correct. I just meant for the validation. Right now if the user passes `--rev <rev> --to-date <date> --from-date <date>`, we'll silently ignore the `--to-date` and `--from-date` since this validation is happening inside the if statement.

I was just asking to move the validation up higher so the above scenario becomes an error.

EvaBardou: Ok, thx for the explanation!

mozci/console/commands/push.py
marco-c: Follow-up filed: https://github.com/mozilla/mozci/issues/547

marco-c (approved):

Issue #539, opened 2021-11-04
Build Docker image


La0:
A new Docker image should be created embedding mozci and its dependencies.

It should be publicly available, and automatically published from CI:
- `latest` from pushes on master branch
- `X.Y.Z` from pushes on tags X.Y.Z (releases)

That docker image would then be used from Taskcluster tasks to run decisions and classifications

La0:
Closed by #546 : the image is now indexed https://community-tc.services.mozilla.com/tasks/index/project.mozci.docker.revision

PR #546, opened 2021-11-08
Add docker image based on alpine

### LINKED ISSUE #545: zstandard is not optional
The zstandard package is marked as an optional dependency, but if I try to install with:

```console
poetry install --no-dev 
```

A later execution of `mozci` will result in an immediate crash:

```console
2021-11-08 17:24:03.153 | WARNING  | mozci.configuration:__init__:130 - Configuration path /root/.config/mozci/config.toml is not a file.
Traceback (most recent call last):
  File "/usr/local/bin/mozci", line 2, in <module>
    from mozci.console.application import cli
  File "/src/mozci/__init__.py", line 2, in <module>
    from mozci.configuration import config  # noqa
  File "/src/mozci/configuration.py", line 187, in <module>
    config = Configuration()
  File "/src/mozci/configuration.py", line 135, in __init__
    self.cache = CustomCacheManager(self._config["cache"])
  File "/src/mozci/configuration.py", line 90, in __init__
    self.register_serializer("compressedpickle", CompressedPickleSerializer())
  File "/src/mozci/util/cache_stores.py", line 256, in __init__
    import zstandard
ModuleNotFoundError: No module named 'zstandard'
```

La0:
Closes #545 & #539 

The image is built using [mozilla/task-boot](https://github.com/mozilla/task-boot/) + Taskcluster dind.

It's currently available as a Taskcluster artifact. We now have 2 choices for further publication:
1. only publish as artifact, and index the task so that other Taskcluster task (on the same instance) can use that image
2. add Docker hub credentials in a secret, and use it through `task-boot push-artifact` in order to publish the image on Dockerhub.

We also need to define a publication strategy:
- push as `latest` on `master` (merges) ?
- push as `X.Y.Z` on tags `X.Y.Z` (releases) ?
- push on specific production or other branches ?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM python:3.9-alpine
+
+# Add mozci source code and setup instructions
+WORKDIR /src
+COPY LICENSE  poetry.lock  pyproject.toml  README.md ./
+COPY mozci /src/mozci/
+
+# Run in a single step to get a small image
+RUN apk add --virtual build gcc libffi-dev musl-dev postgresql-dev && \
+  # Setup latest poetry through pip
+  pip install poetry && \
+  # Install mozci with poetry
+  poetry config virtualenvs.create false && \
+  poetry install --no-dev --no-interaction --no-ansi && \
+  # Cleanup build dependencies
+  apk del build && \
+  rm -rf /tmp/build
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ COPY mozci /src/mozci/
 # Run in a single step to get a small image
 RUN apk add --virtual build gcc libffi-dev musl-dev postgresql-dev && \
   # Setup latest poetry through pip
-  pip install poetry && \
+  pip install --no-cache-dir poetry && \
   # Install mozci with poetry
   poetry config virtualenvs.create false && \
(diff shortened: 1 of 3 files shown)


ahal (approved):
Lgtm!

La0:
The docker image is now indexed and usable on Taskcluster ([demo on last CI run](https://community-tc.services.mozilla.com/tasks/NqaDUL0PQyG6N_rKpQeCxQ/runs/0/logs/live/public/logs/live.log))

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Add mozci source code and setup instructions
+WORKDIR /src
+COPY LICENSE  poetry.lock  pyproject.toml  README.md ./
+COPY mozci /src/mozci/
+
+# Run in a single step to get a small image
+RUN apk add --virtual build gcc libffi-dev musl-dev postgresql-dev && \
+  # Setup latest poetry through pip
+  pip install poetry && \
```
marco-c, line 11: Nit: add `--no-cache-dir` to avoid the pip cache being included in the image

marco-c (approved):

Dockerfile
La0: Done :+1:

Issue #541, opened 2021-11-04
Build cleo command to run Taskcluster decisions for classification


La0:
A new command named `decision-task` should be created and :
1. list all the running pushes from Treeherder api `push > list`
2. create a new children task for each push, using the mozci image and the `classify` command

A `--dry-run` mode should be implemented to run locally, without Taskcluster and see which tasks would be created.

marco-c:
The hard part here is deciding which pushes to analyze.

La0:
The [hook](https://community-tc.services.mozilla.com/hooks/project-mozci/decision-task-testing) is now running every 15 minutes to process the last 15 pushes.

PR #552, opened 2021-11-18
Build simple decision task spawning child tasks on CI

La0:
Refs #541

### DIFF
--- a/mozci/console/application.py
+++ b/mozci/console/application.py
@@ -4,12 +4,14 @@ import sys
 
 from cleo import Application
 
+from mozci.console.commands.decision import DecisionCommand
 from mozci.console.commands.push import PushCommands
 
 
 def cli():
     application = Application()
     application.add(PushCommands())
+    application.add(DecisionCommand())
     application.run()
 
 
--- a/mozci/console/commands/decision.py
+++ b/mozci/console/commands/decision.py
@@ -6,8 +6,7 @@ from datetime import datetime
 import taskcluster
 from cleo import Command
 
-from mozci import data
-from mozci.push import Push
+from mozci.push import Push, make_push_objects
 from mozci.util.memoize import memoized_property
 from mozci.util.taskcluster import get_proxy_queue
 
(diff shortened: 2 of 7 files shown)


mozci/console/application.py
```diff
@@ -4,12 +4,14 @@
...
 from cleo import Application
 
+from mozci.console.commands.decision import DecisionCommand
 from mozci.console.commands.push import PushCommands
 
 
 def cli():
     application = Application()
     application.add(PushCommands())
+    application.add(DecisionCommand())
```
marco-c, line 14: Maybe we want to name this command more explicitly? Something like `trigger-classifications`? Wdyt @ahal?

mozci/console/commands/decision.py
```diff
@@ -0,0 +1,116 @@
...
+        # List pushes
+        pushes = data.handler.get(
+            "pushes",
+            branch=branch,
+        )
+
+        for push in pushes:
+
+            # TODO: detect if a classification is already available
+            # and can be skipped
```
marco-c, line 41: Given our discussion from yesterday, we might not want to do this to avoid having to store state somewhere and instead just rely on the age of a push.
Let's remove the TODO and maybe file it as an "investigate XXX" issue.

mozci/console/commands/decision.py
```diff
@@ -0,0 +1,116 @@
...
+                    }
+                },
+            },
+            "routes": [
+                f"project.mozci.classification.{push.branch}.revision.{push.rev}",
+                f"project.mozci.classification.{push.branch}.push.{push.id}",
+            ],
+            "provisionerId": self.current_task["provisionerId"],
+            "workerType": self.current_task["workerType"],
+        }
```
marco-c, line 112: Do we need to define the group too, or is the new task going to be automatically in the same group as the decision task given the dependency?

mozci/data/sources/treeherder/__init__.py
```diff
@@ -11,6 +11,7 @@
 
 from mozci.data.base import DataSource
 from mozci.errors import ContractNotFilled
+from mozci.push import Push as MozCiPush
```
marco-c, line 14: Won't this create a circular dependency?

marco-c: You could follow the same strategy as `make_push_objects` (get data from the data source, then instantiate Push objects from that data).

marco-c: Actually, we could just extend make_push_objects to be able to either use dates (as it is right now), or number of pushes

mozci/console/commands/decision.py
La0: We do need to define it !

mozci/data/sources/treeherder/__init__.py
La0: Done :+1:

marco-c (approved):
LGTM!

mozci/console/application.py
ahal: Actually I kind of wonder if integrating into the cli at all is the right approach here, since no one will be able to run it anyway.. Maybe this should just be script and then the hook would explicitly run `python mozci/scripts/decision.py` or something like that?

PR #562, opened 2021-11-24
Add a new Cleo command to compare classification results with reality

### LINKED ISSUE #553: Implement a command to evaluate the classification results by comparing them with reality
The command should calculate statistics such as:
- how many pushes that we identified as bad were actually backed-out (true positives);
- how many pushes that we identified as bad were not backed-out (false positives);
- how many pushes that we identified as good were backed-out (false negatives);
- how many pushes that we identified as good were not backed-out (true negatives);
- percentage of pushes we weren't able to identify as good or bad.

EvaBardou:
As we can't retrieve classification results from artifacts for now, the command only displays pushes that were backed-out or not.

### TODO:
- [x] Retrieve classification results
- [x] Output statistics

Closes #553

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -6,7 +6,8 @@ import os
 from cleo import Command
 from tabulate import tabulate
 
-from mozci.push import Push, make_push_objects
+from mozci.push import Push, PushStatus, make_push_objects
+from mozci.task import Task
 
 
 class PushTasksCommand(Command):
@@ -175,6 +176,9 @@ class ClassifyEvalCommand(Command):
         {--rev= : Head revision of the push.}
         {--from-date= : Lower bound of the push range (as a date in yyyy-mm-dd format).}
         {--to-date= : Upper bound of the push range (as a date in yyyy-mm-dd format).}
+        {--medium-confidence=0.8 : If recalculate parameter is set, medium confidence threshold used to classify the regressions.}
+        {--high-confidence=0.9 : If recalculate parameter is set, high confidence threshold used to classify the regressions.}
+        {--recalculate : If set, recalculate the classification instead of fetching an artifact.}
     """
 
     def handle(self):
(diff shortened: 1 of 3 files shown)


marco-c:
Could you add an option to decide between retrieving them from the tasks and recalculating them? The second could be useful when we want to test locally changes to the algorithm.

EvaBardou:
> Could you add an option to decide between retrieving them from the tasks and recalculating them locally? The second could be useful when we want to test locally changes to the algorithm.

Sure, no problem.

mozci/util/taskcluster.py
```diff
@@ -37,16 +38,18 @@ def _handle_artifact(path, response):
     return response
 
 
-def get_artifact_url(task_id, path):
+def get_artifact_url(task_id, path, use_community=False):
```
marco-c, line 41: I'd prefer a `root_url` parameter which is `PRODUCTION_TASKCLUSTER_ROOT_URL` by default, it's more generic

mozci/console/commands/push.py
```diff
@@ -28,6 +29,37 @@ def handle(self):
         self.line(tabulate(table, headers=["Label", "Result"]))
 
 
+def classify_commands_pushes(branch, from_date, to_date, rev):
+    if not (bool(rev) ^ bool(from_date or to_date)):
+        return (
```
marco-c, line 34: Instead of returning the error as a second return value, let's raise an exception and catch it in the caller

mozci/console/commands/push.py
```diff
@@ -28,6 +29,37 @@ def handle(self):
         self.line(tabulate(table, headers=["Label", "Result"]))
 
 
+def classify_commands_pushes(branch, from_date, to_date, rev):
```
marco-c, line 32: Nit: can you add type definitions?

mozci/console/commands/push.py
```diff
@@ -165,14 +167,117 @@ def handle(self):
...
+class ClassifyEvalCommand(Command):
+    """
+    Evaluate the classification results for a given push (or a range of pushes) by comparing them with reality.
+
+    classify-eval
+        {branch=mozilla-central : Branch the push belongs to (e.g autoland, try, etc).}
+        {--rev= : Head revision of the push.}
+        {--from-date= : Lower bound of the push range (as a date in yyyy-mm-dd format).}
+        {--to-date= : Upper bound of the push range (as a date in yyyy-mm-dd format).}
+        {--medium-confidence=0.8 : If recalculate parameter is set, medium confidence threshold used to classify the regressions.}
```
marco-c, line 179: We should print an error if one of these parameters is passed but `recalculate` is not

marco-c:
Looks good overall, just a few minor comments!

marco-c (approved):
LGTM.
I'm not super happy that we have to pass the TC instance around, but I can't think of another solution. @ahal @La0 do you have an alternative idea?

mozci/console/commands/push.py
```diff
@@ -165,14 +175,137 @@ def handle(self):
...
+                    good_non_backedout_pushes.append(push)
+            else:
+                unknown_pushes.append(push)
+
+        if classification_failed:
+            self.line(
+                f"<error>Failed to fetch or recalculate classification for {len(classification_failed)} out of {len(pushes)} pushes.</error>"
+            )
+        self.line(
+            f"{len(bad_backedout_pushes)} out of {len(pushes)} pushes were backed-out by a sheriff and were classify as BAD."
```
ahal, line 285: s/classify/classified

Same for the ones below

ahal (approved):

marco-c (approved):

PR #569, opened 2021-11-30
Add 'is_finalized' property to pushes to control caching

### LINKED ISSUE #566: Add a "is finalized" attribute to pushes to control caching
There are some things we cannot cache, for example the tasks belonging to a push, until a push is totally "finished".
If we cache a push while its tasks are in progress, we will never be able to know if the tasks ended up failing or passing.

We still want to cache things though, or classifications will be too slow.

We should introduce a "is finalized" attribute that is `True` when a push is "reasonably" finished (a push is never actually finished because you could always add new tasks, but we can reasonably assume nobody is going to add new tasks to very old pushes). Then, we should use this attribute to decide if we can cache some things (e.g. the tasks).

We could simply use the age of the push (> 1 day old -> True).

EvaBardou:
Closes #566

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -157,7 +157,7 @@ class Push:
         Returns:
             bool: True if the push is considered finalized (> 1 day old), else False.
         """
-        yesterday = (datetime.now() - timedelta(days=1)).timestamp() * 1000
+        yesterday = (datetime.now() - timedelta(days=1)).timestamp()
         return self.date < yesterday
 
     def create_push(self, push_id):
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -6,6 +6,7 @@ import math
 from argparse import Namespace
 from collections import defaultdict
 from dataclasses import dataclass
+from datetime import datetime, timedelta
 from enum import Enum
 from typing import Dict, Iterator, List, Optional, Set, Tuple, Union
 
@@ -148,6 +149,16 @@ class Push:
 
         self._id = self._hgmo.pushid
         return self._id
+
+    @property
+    def is_finalized(self):
+        """The push is finished or not.
+
+        Returns:
+            bool: True if the push is considered finalized (> 1 day old), else False.
+        """
+        yesterday = (datetime.now() - timedelta(days=1)).timestamp() * 1000
+        return self.date < yesterday
 
     def create_push(self, push_id):
         result = HgRev.load_json_push(self.branch, push_id)
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -336,7 +336,7 @@ class Push:
         # Skip tier tasks greater than the tier passed in config
         tasks = [task for task in tasks if not task.tier or task.tier <= config.tier]
 
-        # Push is supposedly finalized, we can cache the results as tasks should've finish running
+        # Push is supposedly finalized, we can cache the results as tasks should've finished running
         if self.is_finalized:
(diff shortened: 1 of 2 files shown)


marco-c:
We should also use the attribute in other places, e.g. to control caching of tasks.

EvaBardou:
@marco-c WDYM by caching of tasks? What is the data you wish to cache too and in which function it is?

> We should also use the attribute in other places, e.g. to control caching of tasks.

marco-c:
This one: https://github.com/mozilla/mozci/blob/62033292018353ac6c3a136e30c5fc7ac3acaef3/mozci/push.py#L329.

EvaBardou:
@marco-c so I should just revert the automatic caching to a conditional one there, that's it?

marco-c:
Exactly!

mozci/push.py
```diff
@@ -149,6 +150,16 @@ def id(self):
         self._id = self._hgmo.pushid
         return self._id
 
+    @property
+    def is_finalized(self):
```
marco-c, line 154: @ahal any other idea how we can tell if a push is finalized?

mozci/push.py
```diff
@@ -324,15 +336,16 @@ def tasks(self):
...
         tasks = [task for task in tasks if not task.tier or task.tier <= config.tier]
 
-        # Now we can cache the results.
-        # cachy's put() overwrites the value in the cache; add() would only add if its empty
-        config.cache.put(
-            f"{self.push_uuid}/tasks",
-            tasks,
-            config["cache"]["retention"],
-        )
+        # Push is supposedly finalized, we can cache the results as tasks should've finish running
```
marco-c, line 339:
```suggestion
        # Push is supposedly finalized, we can cache the results as tasks should've finished running
```

mozci/push.py
```diff
@@ -324,15 +336,16 @@ def tasks(self):
...
+        if self.is_finalized:
+            # cachy's put() overwrites the value in the cache; add() would only add if its empty
+            config.cache.put(
+                f"{self.push_uuid}/tasks",
+                tasks,
+                config["cache"]["retention"],
+            )
```
marco-c, line 346: A future optimization that we could do is to cache tasks, and then on the next run only used the cached tasks that were finished.
This way, we can avoid `Gathering test groups` for tasks that were already finished in the previous run and just do it for newly finished tasks.

marco-c: I filed #570 as a follow-up.

mozci/push.py
```diff
@@ -1094,7 +1107,25 @@ def classify(
         - good push: when there are only intermittent failures
         - unknown state: when other tasks are failing
         """
-        regressions = self.classify_regressions(confidence_medium, confidence_high)
+        cache_path = f"{self.push_uuid}/classify_group_tasks/classified_regressions"
```
marco-c, line 1110: Maybe we should avoid to cache the classify results completely:
- when we run the command to analyze pushes for debugging, we don't want the cache, as we want to experiment with different values and variations of the algorithm;
- when we run the command to analyze pushes at "real time", the cache is not useful because the push is definitely not finalized.

So in the end we'd basically never benefit from this specific caching.

EvaBardou: Ok, so I should remove it then?

marco-c: Yes, thanks

PR #571, opened 2021-12-01
Use task errors when listing failing tasks

### LINKED ISSUE #563: Some "intermittent" classifications have empty list of tasks
See for example https://community-tc.services.mozilla.com/tasks/index/project.mozci.classification.mozilla-central.push/989898.

I'm attaching the classifications JSON file.
[classification.zip](https://github.com/mozilla/mozci/files/7604152/classification.zip)
There are some groups under "intermittents" which have an empty list of tasks, e.g. accessible/tests/mochitest/attributes/a11y.ini.

La0:
Closes #563 

The current `failing_tasks` algorithm only uses completed tasks with results: this totally skips the failed tasks with errors.

I made a bunch of changes here:
- `failing_tasks` now also checks if the task has any error
- task errors are loaded at the same time than task results
- missing test results are marked as failed
- fixed a bug on the task errors output (validx was not OK with current implementation)

### DIFF
--- a/mozci/data/sources/artifact/__init__.py
+++ b/mozci/data/sources/artifact/__init__.py
@@ -79,4 +79,4 @@ class ErrorSummarySource(DataSource):
     def run_test_task_errors(self, task):
         if task.id not in self.TASK_ERRORS:
             self._load_errorsummary(task.id)
-        return self.TASK_ERRORS.pop(task.id, {})
+        return self.TASK_ERRORS.pop(task.id, [])
--- a/mozci/data/sources/artifact/__init__.py
+++ b/mozci/data/sources/artifact/__init__.py
@@ -44,6 +44,11 @@ class ErrorSummarySource(DataSource):
             if line
         )
 
+        def _add_error(task_id, message):
+            if task_id not in self.TASK_ERRORS:
+                self.TASK_ERRORS[task_id] = []
+            self.TASK_ERRORS[task_id].append(message)
+
         for line in lines:
             if line["action"] == "test_groups":
                 groups |= set(line["groups"]) - {"default"}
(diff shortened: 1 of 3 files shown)


La0:
And [here](https://gist.github.com/La0/9758c0c42723d342d436fa9558bbb250) is the output for the push mentionned in the issue

marco-c:
> And [here](https://gist.github.com/La0/9758c0c42723d342d436fa9558bbb250) is the output for the push mentionned in the issue

There are still some intermittent groups without tasks associated to them, e.g. `"testing/web-platform/tests/css/css-counter-styles/oriya": [],`.

mozci/data/sources/artifact/__init__.py
```diff
@@ -55,21 +57,24 @@ def _load_errorsummary(self, task_id) -> None:
...
+            if result != "SKIP"
+        }
+
         missing_groups = groups - set(group_results)
         if len(missing_groups) > 0:
             logger.error(
                 f"Some groups in {task_id} are missing results: {missing_groups}"
             )
+            for group in missing_groups:
+                self.TASK_GROUPS[task_id][group] = False
```
marco-c, line 74: We shouldn't assume that a missing result means a failing test, in most cases if a group is missing it means it didn't run at all.

mozci/task.py
```diff
@@ -413,12 +415,14 @@ def status(self):
...
     def failing_tasks(self):
         # List all tasks with some test results failing for that group
+        # or with misbehaving tasks (early failures or missing results)
         return [
             task
             for task in self.tasks
             if any(
                 not result.ok and result.group == self.name for result in task.results
             )
+            or len(task.errors) > 0
```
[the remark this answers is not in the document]
marco-c: For example, OVuTZ20wTxa60Rd29DcfLA failed.
The test that failed was browser/base/content/test/general/browser_save_video_frame.js, belonging to group "browser/base/content/test/general/browser.ini".
The error is the last line of https://firefoxci.taskcluster-artifacts.net/OVuTZ20wTxa60Rd29DcfLA/0/public/test_info/mochitest-browser-chrome_errorsummary.log, and it isn't clear from the data which group the error belongs to.
If we applied this change, we would consider all groups in OVuTZ20wTxa60Rd29DcfLA as failing, even though actually only one of them failed.

mozci/data/sources/artifact/__init__.py
marco-c: Same reasoning as https://github.com/mozilla/mozci/pull/571#discussion_r760937698

ahal (changes requested):
Generally looks good but Marco's comment needs to be fixed, so removing from my queue.

La0:
Not the right approach, closing

Issue #564, opened 2021-11-25
Implement a shared cache between the classification tasks


marco-c:
Right now each classification task is totally separate, which means they are retrieving the same data multiple times.
We should introduce a shared volume between the classification tasks to avoid that.

webbdays:
@marco-c 
I would like to work on this issue.
Please provide code location for that.

webbdays:
@marco-c @La0 

Please provide the example for classification task. What are the things on which the classification task is performed ?

marco-c:
@webbdays this is not an easy issue to solve. @La0 is working on it.

webbdays:
Ok

La0:
The cache is now enabled ([demo](https://community-tc.services.mozilla.com/tasks/groups/dWDbOHzHSdiGmXGsFNUyrw))

PR #573, opened 2021-12-02
Use Taskcluster secret as configuration source to use cache

### LINKED ISSUE #572: Move get_s3_credentials in taskcluster module
I just noticed that `mozci.util.cache_stores.get_s3_credentials` (and its dependant method `get_taskcluster_options`) could be moved into the `mozci.util.taskcluster` module where there is also a method to use taskcluster proxy.

La0:
Closes #572 & #564 

A few things are happening here:
1. a docker-worker cache is enabled for children tasks (see scope request https://github.com/mozilla/community-tc-config/pull/455 )
2.  in order to use that cache, we need to configure mozci in Taskcluster. The best way to do that is to use a Taskcluster secret
3. so the `Configuration` class uses an optional environment variable `TASKCLUSTER_CONFIG_SECRET` that is also set on the children tasks (the decision task does not need configuration).
4. then it's fairly immediate: any child task will use that configuration and merge it as it would from a file. So if we set the correct configuration in `project/mozci/testing` the shared cache will be used

Maybe we should add some debugging output regarding the cache configuration to make sure we are using correctly ?

### DIFF
--- a/mozci/configuration.py
+++ b/mozci/configuration.py
@@ -4,9 +4,11 @@ import os
 from collections.abc import Mapping
 from pathlib import Path
 
+import taskcluster
 from appdirs import user_config_dir
 from cachy import CacheManager
 from loguru import logger
+from taskcluster.helper import load_secrets
 from tomlkit import parse
 
 from mozci.util.cache_stores import (
@@ -16,6 +18,7 @@ from mozci.util.cache_stores import (
     S3Store,
     SeededFileStore,
 )
+from mozci.util.taskcluster import get_taskcluster_options
 
 
 def merge_to(source, dest):
@@ -97,6 +100,7 @@ class CustomCacheManager(CacheManager):
 
 class Configuration(Mapping):
     DEFAULT_CONFIG_PATH = Path(user_config_dir("mozci")) / "config.toml"
+    TASKCLUSTER_CONFIG_SECRET = os.environ.get("TASKCLUSTER_CONFIG_SECRET")
     DEFAULTS = {
         "merge": {
             "cache": {"retention": 1440},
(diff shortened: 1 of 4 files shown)


La0:
I was able to test the secrets loading from firefox-tc with the following conf:

```console
#!/bin/bash
export TASKCLUSTER_ROOT_URL="https://firefox-ci-tc.services.mozilla.com"
export TASKCLUSTER_CLIENT_ID="mozilla-auth0/ad|Mozilla-LDAP|XXX/YYY"
export TASKCLUSTER_ACCESS_TOKEN="...."
export TASKCLUSTER_CONFIG_SECRET="path/to/a/secret"

mozci push classify --rev=c73bcc465e0c2bce7debb0a86277e2dcb27444e4
```

@marco-c Can you setup the secret `project/mozci/testing` (on community-tc) with the following content:

```yaml
---
mozci:
  verbose: 1

  cache:
    retention: 1209600 # 2 weeks

    stores:
      file:
        driver: file
        path: /cache
```

mozci/configuration.py
```diff
@@ -79,6 +82,10 @@ def __init__(self, cache_config):
         super_config.setdefault("stores", {"null": {"driver": "null"}})
         super(CustomCacheManager, self).__init__(super_config)
 
+        for store, conf in self._config["stores"].items():
+            if store != "null":
+                logger.debug(f"Active cache store {store} with conf {conf}")
```
marco-c, line 87: I guess you were planning to remove this?

marco-c (approved):

marco-c:
@marco-c Can you setup the secret project/mozci/testing (on community-tc) with the following content:

Done

mozci/configuration.py
La0: Well no, I mean that's up to debate.

My idea here is to leave this enabled for now and see if the log line is displayed on Taskcluster.

It may be pretty hard to be sure that the cache is working otherwise  as we don't have access to the docker-worker cache directly.

PR #580, opened 2021-12-16
Cache tasks more often than only when a push is finalized

### LINKED ISSUE #570: Cache tasks more often than only when a push is finalized
Right now (that is, after #569), we are only caching tasks if a push is finalized.

We could instead cache all tasks and, in the next run, re-retrieve only tasks that in the previous run weren't finished.

This wouldn't change anything for the [Gather data about tasks that ran on a given push](https://github.com/mozilla/mozci/blob/62033292018353ac6c3a136e30c5fc7ac3acaef3/mozci/push.py#L274) and [Gather task classifications](https://github.com/mozilla/mozci/blob/62033292018353ac6c3a136e30c5fc7ac3acaef3/mozci/push.py#L283) steps, but it would reduce the need for [Gather group data](https://github.com/mozilla/mozci/blob/62033292018353ac6c3a136e30c5fc7ac3acaef3/mozci/push.py#L297) (which is also the slowest part, so that's good!).

EvaBardou:
Closes #570


mozci/push.py
```diff
@@ -298,7 +306,11 @@ def tasks(self):
                 "push_tasks_classifications", branch=self.branch, rev=self.rev
             )
             for task in tasks:
-                if task["id"] in classifications:
+                # No need to gather task classifications for a completed task that was already cached
+                if (
+                    task["id"] not in completed_cached_tasks
+                    and task["id"] in classifications
+                ):
```
marco-c, line 313: Classifications might actually change (sheriffs can classify a few hours after a failure), so let's keep the update no matter what (it's not expensive anyway, the expensive part is the `push_tasks_classifications` query above, which we have to do anyway).

marco-c (approved):
LGTM (after the change suggested above), can you add a couple of tests or file a follow-up issue to add them?

marco-c:
It looks like tests are failing.

EvaBardou:
> It looks like tests are failing.

Yes i'm working on it

marco-c (approved):

PR #582, opened 2021-12-16
Add a CLI command to analyze execution times of classification hook

La0:
With this PR you can run `mozci push perf` to generate a CSV file with data about ALL classification tasks from a mozci hook.

Columns are :

-  branch
-  revision
-  push ID
-  task ID
-  task creation date
-  time taken by all runs in seconds

This allow a developer to analyze the execution time taken by classification for a given push (or range of pushes) over a period of time.

The command uses cache to avoid re-requesting completed tasks


La0:
WDYT @marco-c & @ahal : I've now used it several times to build the CSV file to produce stats & analysis. Should we get this merged or just close the PR ?

marco-c:
I think we should get it merged, it can be useful in the future too.

mozci/console/commands/push.py
```diff
@@ -305,14 +314,160 @@ def handle(self):
...
+
+    def list_classification_tasks(self, group_id):
+        cache_key = f"perf/task_group/{group_id}"
+
+        # Check cache
+        tasks = config.cache.get(cache_key)
+        if tasks is None:
+            queue = taskcluster.Queue(get_taskcluster_options())
+            try:
+                tasks = queue.listTaskGroup(group_id).get("tasks", [])
```
marco-c, line 429: Do we need pagination or is it automatically handled by the taskcluster library?

marco-c (approved):

mozci/console/commands/push.py
La0: Argh pagination is not handled automatically

La0: Just added support for pagination

marco-c (approved):

PR #596, opened 2022-01-12
Implement ability to perform retriggers on a push (Fixes #508)

### LINKED ISSUE #508: Implement ability to perform retriggers and backfills on a push
In order to facilitate certain sheriff workflows we'll need to be able to run actions on a push. Things like re-triggering and backfills.

This depends on #507

chichi012:
Fixes #508

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -4,12 +4,14 @@ from __future__ import annotations
 import os
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
+from datetime import datetime
 from enum import Enum
 from inspect import signature
 from statistics import median
 from typing import Dict, List, Optional
 
 import requests
+import taskcluster
 from loguru import logger
 
 from mozci import data
@@ -17,8 +19,10 @@ from mozci.errors import ArtifactNotFound, TaskNotFound
 from mozci.util.memoize import memoized_property
 from mozci.util.taskcluster import (
     PRODUCTION_TASKCLUSTER_ROOT_URL,
+    create_task,
     find_task_id,
     get_artifact,
+    get_task,
     list_artifacts,
 )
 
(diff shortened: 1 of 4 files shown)


mozci/task.py
```diff
@@ -218,6 +222,14 @@ def to_json(self):
         sig = signature(self.__init__)
         return {k: v for k, v in self.__dict__.items() if k in sig.parameters}
 
+    def retrigger(self):
+        new_task_id = taskcluster.slugId()
```
ahal, line 226: I think we'll need to implement this `_should_retrigger` check here too:
https://searchfox.org/mozilla-central/source/taskcluster/gecko_taskgraph/actions/retrigger.py#160

Some tasks could cause problems if they get retriggered, so they have a way of declaring that they do not. This check acts as a safeguard to prevent accidentally retriggering them when we shouldn't.

mozci/task.py
```diff
@@ -218,6 +222,14 @@ def to_json(self):
         sig = signature(self.__init__)
         return {k: v for k, v in self.__dict__.items() if k in sig.parameters}
 
+    def retrigger(self):
```
ahal, line 225: Please add a docstring for this function that defines what it does, this will ensure it gets listed in the `mozci` API documentation :)

ahal (changes requested):

ahal:
Would you also mind adding a test to `test_task.py`?

Obviously we won't want the test to *actually* retrigger tasks, but you can use the [responses](https://github.com/getsentry/responses) library (used in other places in that file already) to mock out the HTTP requests. If you need a hand setting up the proper response for your test feel free to ping me on Matrix.

jmaher (approved):
Thanks Chidinma, this is really great to get in place.  Looking forward to using this in the coming weeks.

PR #598, opened 2022-01-13
Add a parameter to get regressions without assumptions about backouts

### LINKED ISSUE #597: Ensure classification API doesn't ignore failures on pushes that are not backed-out
When we gather historical data, we can rely on the back-out state of the pushes. That is, we can assume all failures on a push which was not backed-out are intermittent.
When we are classifying pushes in real time instead, we can't rely on that because we can't look into the future and see if the push will be backed-out.

We have to double check that our logic relying on back-out information is not used when classifying pushes in real time.

marco-c:
Fixes #597

### DIFF
--- a/mozci/configuration.py
+++ b/mozci/configuration.py
@@ -119,6 +119,7 @@ class Configuration(Mapping):
             ],
             "verbose": 0,
             "tier": 2,
+            "historical_analysis": True,
         },
     }
 
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -794,7 +794,11 @@ class Push:
         """
         logger.debug(f"Retrieving candidate regressions for {self.rev}...")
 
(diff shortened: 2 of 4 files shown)


marco-c:
I turned it to draft because I was thinking to maybe move it to a parameter instead of a configuration variable, so it's possible to write code that can do both historical and not historical analyses.
I then realized we can make this change to a parameter in the future if we need to.

ahal (approved):
This seems fine to me! I do think a parameter might be better in this case, but it's not a big deal to leave as a config option.

Maybe you could update the docs though:
https://mozci.readthedocs.io/en/latest/configuration.html#list-of-options

marco-c:
I've decided to go ahead and move to a parameter instead.

ahal (approved):

PR #600, opened 2022-01-17
Enhance command push classify-eval

La0:
- Add more logging:
   - around errors
   - on recalculate normal flow
   - on Taskcluster task fetch normal flow
- Refactor the output & inner triage of push classification so it can generate both:
   - same summary output
   - a CSV output with more detailed informations (tracking exceptions)

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -13,6 +13,7 @@ from tabulate import tabulate
 from taskcluster.exceptions import TaskclusterRestFailure
 
 from mozci import config
+from mozci.errors import TaskNotFound
 from mozci.push import Push, PushStatus, make_push_objects
 from mozci.task import Task
 from mozci.util.taskcluster import (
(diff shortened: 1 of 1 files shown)


mozci/console/commands/push.py
```diff
@@ -232,23 +235,23 @@ def handle(self):
...
                 try:
                     classification, _ = push.classify(
                         confidence_medium=medium_conf, confidence_high=high_conf
                     )
-                except Exception:
-                    classification_failed.append(push)
-                    continue
+                    self.line(
+                        f"<comment>Classified {branch} {push.rev} : {classification.name}</comment>"
+                    )
```
marco-c, line 249: Might be too noisy if we run an analysis on a month worth of pushes

mozci/console/commands/push.py
```diff
@@ -260,48 +263,94 @@ def handle(self):
...
-                if push.backedout:
-                    unknown_backedout_pushes.append(push)
-                else:
-                    unknown_non_backedout_pushes.append(push)
+                    self.classifications[push] = classification = PushStatus[
+                        artifact["push"]["classification"]
+                    ]
+                    self.line(
+                        f"<comment>Found classification {branch} {push.rev} : {classification.name}</comment>"
+                    )
```
marco-c, line 271: Same here

mozci/console/commands/push.py
```diff
@@ -269,48 +276,96 @@ def handle(self):
...
+        self.log_pushes(PushStatus.UNKNOWN, True)
+
+        output = self.option("output")
+        if output:
+            # Build stats for CSV
+            with open(output, "w") as csvfile:
+                writer = csv.DictWriter(
+                    csvfile,
+                    fieldnames=[
+                        "push",
```
marco-c, line 319: We could remove this as the `revision` in a branch already uniquely identifies a push.

marco-c (approved):
LGTM

mozci/console/commands/push.py
La0: Fixed in https://github.com/mozilla/mozci/pull/600/commits/e6d66b06884df15cc6bbd8a48b216b108c9af184

PR #601, opened 2022-01-18
Speedup classification by avoiding to load each push information

La0:
The classification algorithm uses the method `Push.bustage_fixed_by` that trigger an HGMO query for each push, just to get their related Bugzilla bug ID.

This patch uses the `full=1` parameter on HGMO json-pushes method in order to get the description of each revision, and thus their Bugzilla bug ID.

Here is a benchmark running on a random recent autoland revision (command `time mozci push classify-eval --rev=808ac93bdd9847272087ba981173bc877cff0425 --recalculate`)

|            | master    | parse-bug-description |
|------------|-----------|-----------------------|
| cold cache | 1m35,881s | 0m8,662s              |
| warm cache | 0m15,875s | 0m6,002s              |

On another revision which triggers good `time mozci push classify-eval --rev=9c1b66dcaac38344a227cc6cdf3e29272b487c98 --recalculate`, the results are evern better:
- cold cache on master is 2m38,573s
- warm cache on this branch is 0m3,063s

The results stay the same on both branches

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -30,7 +30,7 @@ from mozci.task import (
     TestTask,
     get_configuration_from_label,
 )
-from mozci.util.hgmo import HgRev
+from mozci.util.hgmo import HgRev, parse_bugs
 from mozci.util.memoize import memoize, memoized_property
 
 BASE_INDEX = "gecko.v2.{branch}.revision.{rev}"
@@ -84,20 +84,10 @@ class Push:
             and all(map(lambda r: isinstance(r, dict), revs))
         ):
             # We should get detailed revision objects here
-            def _parse_bug(rev):
-                """Extract Bugzilla bug id from a description in a revision payload"""
-                desc = rev.get("desc")
-                if not desc:
-                    return
-
-                match = REGEX_REV.match(desc)
-                if match is None:
-                    return
-
-                return int(match.group(1))
-
-            # Parse bugs if any
-            self._bugs = set(filter(None, [_parse_bug(rev) for rev in revs]))
+            # and get the list of Bugzilla Bug Ids from the description
+            self._bugs = set(
+                itertools.chain(*[parse_bugs(rev.get("desc", "")) for rev in revs])
+            )
 
             self._revs = [r["node"] for r in revs]
             head_revision = self._revs[0]
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -74,18 +74,20 @@ class Push:
             self._revs = None
             head_revision = revs
             self._bugs = None
-        elif (
-            isinstance(revs, list)
-            and len(revs) > 0
-            and all(map(lambda r: isinstance(r, dict), revs))
-        ):
-            # We should get detailed revision objects here
-            # and get the list of Bugzilla Bug Ids from the description
-            self._bugs = set(
-                itertools.chain(*[parse_bugs(rev.get("desc", "")) for rev in revs])
-            )
+        elif isinstance(revs, list) and len(revs) > 0:
+            if all(map(lambda r: isinstance(r, dict), revs)):
+                # We should get detailed revision objects here
+                # and get the list of Bugzilla Bug Ids from the description
+                self._bugs = set(
+                    itertools.chain(*[parse_bugs(rev.get("desc", "")) for rev in revs])
+                )
 
-            self._revs = [r["node"] for r in revs]
+                self._revs = [r["node"] for r in revs]
+            else:
+                # Support list of changeset IDs
+                self._bugs = None
+                self._revs = revs
+
             head_revision = self._revs[0]
         else:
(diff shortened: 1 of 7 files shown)


mozci/push.py
```diff
@@ -70,10 +74,35 @@ class Push:
...
             self._revs = None
             revs = [revs]
+            self._bugs = None
         else:
-            self._revs = revs
+
+            # We may get detailed revision objects here
+            if all(map(lambda r: isinstance(r, dict), revs)):
+
+                def _parse_bug(rev):
```
marco-c, line 86: You can use https://hg.mozilla.org/hgcustom/version-control-tools/file/9822fcf4b1178d219b7d7a386dda02a11facf55b/pylib/mozautomation/mozautomation/commitparser.py#l97 to have exactly the same behavior as before.

Unrelated question: can you move the function to the top-level?

mozci/util/hgmo.py
```diff
@@ -55,9 +56,14 @@ def _get_and_cache_pushes(branch: str, url: str) -> List[HgPush]:
 
     @staticmethod
     def load_json_pushes_between_ids(
-        branch: str, push_id_start: int, push_id_end: int
+        branch: str, push_id_start: int, push_id_end: int, use_full_format=False
```
marco-c, line 59: What if we *always* use the full format?

marco-c: After all, when looking for regressions we are always calling bustage_fixed_by, so in the end all pushes will be created with this information.

ahal: Seems like a tradeoff of a larger response vs potentially preventing future queries? Sounds worth doing to me.

marco-c: Let's do it then, it also allows simplifying the code a bit.

marco-c:
It might be worth making the hgmo properties try to get their data from json-pushes first, so we're going to less frequently hit json-automationrelevance. We can file a follow-up for that if you prefer (it's a minor perf improvement compared to this PR).

mozci/util/hgmo.py
La0: Done

mozci/push.py
La0: Done

mozci/push.py
```diff
@@ -42,6 +43,9 @@
 FAILURE_CLASSES = ("not classified", "fixed by commit")
 
 
+REGEX_REV = re.compile(r"bug (\d+)", flags=re.IGNORECASE)
```
marco-c, line 46: You're not using this anymore now

mozci/push.py
```diff
@@ -70,20 +74,35 @@ class Push:
...
+            # and get the list of Bugzilla Bug Ids from the description
+            self._bugs = set(
+                itertools.chain(*[parse_bugs(rev.get("desc", "")) for rev in revs])
+            )
+
+            self._revs = [r["node"] for r in revs]
+            head_revision = self._revs[0]
         else:
-            self._revs = revs
+            raise NotImplementedError(f"Cannot process revisions: {revs}")
```
marco-c, line 95: Why not keeping support for passing a list of str?

mozci/push.py
```diff
@@ -633,11 +656,15 @@ def _iterate_children(self, max_depth=None):
...
+            # We load ALL available information for the pushes
+            # to avoid later queries for bugs in bustage_fixed_by
             next_id = other.id + 1
             if next_id not in HgRev.JSON_PUSHES_CACHE:
                 depth = max_depth or MAX_DEPTH
                 HgRev.load_json_pushes_between_ids(
-                    self.branch, other.id, next_id + depth - i
+                    self.branch,
+                    other.id,
+                    next_id + depth - i,
```
marco-c, line 667: nit: this can go back to how it was before since nothing changed

mozci/util/hgmo.py
```diff
@@ -12,14 +13,56 @@
 
 HgPush = NewType("HgPush", Dict[str, Any])
 
+RE_SOURCE_REPO = re.compile(r"^Source-Repo: (https?:\/\/.*)$", re.MULTILINE)
```
marco-c, line 16: Could you mention where the code comes from? So in the future if there are changes to the original code, we know that we can apply them here too

marco-c:
Here or as a follow-up, we can optimize `make_push_objects` to also set the bugs property of the push, given that `push_revisions` is returning the desc now.

mozci/push.py
La0: Because you mentionned using only full payloads (or a single revision to kickstart the push).

Is there a use case for a list of strings now ?

marco-c: Just as an alternative to single revision, if an external user wants to instantiate a Push directly

marco-c (approved):
Great improvement!

PR #619, opened 2022-01-30
Extend the classify command to perform retriggers on unknown tasks (Fixes #534)

### LINKED ISSUE #534: Extend the classify command to perform retriggers and backfills when it needs more information
The algorithm behind the classify command isn't always able to give a decision between "good" or "bad" and sometimes needs more information in order to be sure.

When a situation like this happens and the `classify` command returns "unknown", it should automatically identify and perform the set of actions (retriggers and/or backfills) that would be necessary to gather enough information.

Depends on https://github.com/mozilla/mozci/issues/504 and https://github.com/mozilla/mozci/issues/508.

chichi012:
Fixes #534 

Assumption:
- `regressions.unknown` returns a map of the manifest name mapped to a list of unknown tasks within that manifest.

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -16,7 +16,7 @@ from taskcluster.exceptions import TaskclusterRestFailure
 
 from mozci import config
 from mozci.errors import SourcesNotFound, TaskNotFound
-from mozci.push import Push, PushStatus, Regressions, make_push_objects
+from mozci.push import Push, PushStatus, Regressions, make_push_objects, retrigger
 from mozci.task import Task, TestTask
 from mozci.util.taskcluster import (
     COMMUNITY_TASKCLUSTER_ROOT_URL,
@@ -128,6 +128,7 @@ class ClassifyCommand(Command):
         {--output= : Path towards a directory to save a JSON file containing classification and regressions details in.}
         {--show-intermittents : If set, print tasks that should be marked as intermittent.}
         {--environment=testing : Environment in which the analysis is running (testing, production, ...)}
+        {--retrigger-unknown : If set, retrigger UNKNOWN regressions}
     """
 
     def handle(self) -> None:
@@ -155,6 +156,7 @@ class ClassifyCommand(Command):
             self.line("<error>Provided --high-confidence should be a float.</error>")
             return
 
+        retrigger_unknown = True if self.option("retrigger-unknown") else False
         output = self.option("output")
         if output and not os.path.isdir(output):
             os.makedirs(output)
@@ -168,6 +170,9 @@ class ClassifyCommand(Command):
                     intermittent_confidence_threshold=medium_conf,
                     real_confidence_threshold=high_conf,
                 )
+                if retrigger_unknown:
+                    for _, tasks in regressions.unknown.items():
+                        retrigger(tasks=tasks, repeat_retrigger=1)
                 self.line(
                     f"Push associated with the head revision {push.rev} on "
                     f"the branch {self.branch} is classified as {classification.name}"
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -1326,6 +1326,13 @@ def make_push_objects(**kwargs):
     return pushes
 
 
+def retrigger(tasks: List[TestTask], repeat_retrigger: int=1) -> None:
+    """Utility function that retrigger a list of task "repeat_retrigger" times."""
+    for task in tasks:
+        for _ in range(0, repeat_retrigger):
+            task.retrigger()
+
+
 def make_summary_objects(from_date, to_date, branch, type):
     """Returns a list of summary objects matching the parameters.
 
--- a/tests/test_push.py
+++ b/tests/test_push.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import json
+import re
 from itertools import count
 
 import pytest
@@ -12,7 +14,7 @@ from mozci.errors import (
     PushNotFound,
     SourcesNotFound,
 )
-from mozci.push import Push, PushStatus, Regressions
+from mozci.push import Push, PushStatus, Regressions, retrigger
 from mozci.task import GroupResult, GroupSummary, Status, Task, TestTask
 from mozci.util.hgmo import HgRev
 from mozci.util.taskcluster import (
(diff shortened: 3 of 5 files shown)


chichi012:
@jmaher @ahal Kindly help review when you can. I plan to write tests for my changes when cleared.

mozci/console/commands/push.py
```diff
@@ -87,6 +87,7 @@ class ClassifyCommand(Command):
         {--high-confidence=0.9 : High confidence threshold used to classify the regressions.}
         {--output= : Path towards a directory to save a JSON file containing classification and regressions details in.}
         {--show-intermittents : If set, print tasks that should be marked as intermittent.}
+        {--retrigger= : If set, retrigger UNKNOWN regressions}
```
jmaher, line 90: is retrigger a boolean, it probably doesn't need the `=` after it.

mozci/push.py
```diff
@@ -1170,7 +1173,11 @@ def classify(
         if len(regressions.unknown) == 0 and len(regressions.intermittent) >= 0:
             return PushStatus.GOOD, regressions
 
-        # Fallback to unknown
+        # Fallback to unknown failures and retrigger each unknown task
+        if should_retrigger:
+            for _, tasks in regressions.unknown.items():
+                for task in tasks:
+                    task.retrigger()
```
jmaher, line 1180: this most likely retriggers the task a single time, typically we will need more data, I would vote for 4 times (when added to the original task, that would equal 5 total tasks to determine status).  The problem is that if we already have a retrigger and both failed for different reasons, we would have a large number of extra tasks that are not needed.

Think about how to extend this, but for the purposes of this review and PR, lets focus on a single retrigger.

jmaher (changes requested):
only 1 issue I see with the `=` after `--retrigger=`

mozci/push.py
chichi012: @jmaher I can extend the code to call retrigger x4 if that is our end goal. Do we want to get that in now?

jmaher: can you make it a variable for number of retriggers, and then we can set it now to a default value of 1.

chichi012: Sounds good. Updated the PR

jmaher:
I like the changes, I need to dig in to the test to see if there is reasonable coverage- most likely yes.

jmaher (changes requested):
is there a way to add any tests to push classification for this?
https://github.com/mozilla/mozci/blob/c130bb7055d77b9b551c348f9d85f535ee169e39/tests/test_push.py#L1061

I think ensuring a known set of tasks that are unknown are retriggered via expected.  Also making sure with retrigger=False (default) that no tasks are retriggered.

Outside of that I don't think there is anything needed to land this.

chichi012:
@jmaher @ahal for the test in `test_push.py`, I don't know why `responses.calls` is empty even when I know that HTTP requests were made.

chichi012:
@jmaher @ahal I believe it's ready now. I've been able to resolve the issues. The problem was the different test cases I was using. It tried checking if a retrigger or classification happened even when I didn’t put any tasks in them except one. But it has been rectified now.

mozci/console/commands/push.py
```diff
@@ -87,6 +87,7 @@ class ClassifyCommand(Command):
         {--high-confidence=0.9 : High confidence threshold used to classify the regressions.}
         {--output= : Path towards a directory to save a JSON file containing classification and regressions details in.}
         {--show-intermittents : If set, print tasks that should be marked as intermittent.}
+        {--retrigger : If set, retrigger UNKNOWN regressions}
```
ahal, line 90: Marco and I had chatted about whether or not this would be better as a cli flag, or as a brand new command. I think I'd mildly prefer a new command, but if Marco / Bastien are ok with the flag here, then that's fine with me too.

If we do go the new command route, I'd call it something like `retrigger-unclassified`, and it would basically just do the exact same thing as this current patch does, except it wouldn't dump the classification results to stdout.

ahal (commented):
Awesome, this looks great to me!

Since this intersects with the auto classification project, I'm flagging Marco and Bastien for review too.

mozci/console/commands/push.py
chichi012: Oh, this sounds great @ahal. Do we do it now, or merge and refactor the code later, or wait for Marco/Bastien's review?

jmaher:
wait for a review from either marco or bastien.

mozci/console/commands/push.py
marco-c: I don't have a preference between the two options (flag vs new command), either wfm.

mozci/console/commands/push.py
```diff
@@ -124,7 +126,9 @@ def handle(self):
         for push in pushes:
             try:
                 classification, regressions = push.classify(
-                    confidence_medium=medium_conf, confidence_high=high_conf
+                    confidence_medium=medium_conf,
+                    confidence_high=high_conf,
```
marco-c, line 130: The names of the arguments changed, can you rebase?

mozci/push.py
```diff
@@ -1151,7 +1151,10 @@ def _map_failing_tasks(groups):
         )
 
     def classify(
-        self, confidence_medium: float = 0.8, confidence_high: float = 0.9
+        self,
+        confidence_medium: float = 0.8,
+        confidence_high: float = 0.9,
+        should_retrigger: bool = False,
```
marco-c, line 1157: Even if we implement the retriggering as a flag of the classify command, I'd write two separate functions, `classify` and `retrigger`.
`retrigger` could accept a list (or iterable) of tasks as input, and the caller will first call `classify` and then call `retrigger` with the list (or iterable) of unknown tasks.

tests/test_push.py
```diff
@@ -1262,6 +1263,107 @@ def test_classify_almost_bad_push(
...
+        r"https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/*"
+    )
+
+    responses.add(responses.PUT, create_new_task_url_matcher, status=200)
+
+    push.classify(should_retrigger=True)
+
+    # verify last call was to create a new task
+    assert (len(responses.calls) != 0 and responses.calls[-1].request.method == "PUT") == expected_result
+    assert (len(responses.calls) != 0 and bool(create_new_task_url_matcher.match(responses.calls[-1].request.url))) == expected_result
```
marco-c, line 1365: Instead of just checking that we retrigger, it might be worth checking that we retrigger the right things

mozci/push.py
```diff
@@ -1170,7 +1173,13 @@ def classify(
         if len(regressions.unknown) == 0 and len(regressions.intermittent) >= 0:
             return PushStatus.GOOD, regressions
 
-        # Fallback to unknown
+        # Fallback to unknown failures and retrigger each unknown task
+        retrigger_repeat_count = 1
```
marco-c, line 1177: This should be an option

marco-c: I mean, a parameter of the function

tests/test_task.py
```diff
@@ -77,7 +77,7 @@ def test_retrigger_should_retrigger(responses, create_task):
     responses.add(
         responses.GET,
         "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/0",
-        json={"payload": {}, "tags": {"retrigger": "true"}},
+        json={"payload": {}, "tags": {"retrigger": "true", "label": "test_retrigger"}},
```
marco-c, line 80: Nit: why is this needed?

marco-c:
Note: retriggering might not always be the right action, depending on why the task is in the list of unknown. For example, another potential action is "backfilling".
We should choose the best action to perform depending on the reason why a task is in the unknown list.

For now this is OK though, we can implement the "best action" decision in a follow-up.

ahal:
> Note: retriggering might not always be the right action, depending on why the task is in the list of unknown. For example, another potential action is "backfilling". We should choose the best action to perform depending on the reason why a task is in the unknown list.
> 
> For now this is OK though, we can implement the "best action" decision in a follow-up.

Agreed, this sounds like a good follow-up though.

tests/test_task.py
chichi012: We needed it because we log the label of the task that is being retriggered.

mozci/console/commands/push.py
chichi012: okay done

mozci/push.py
chichi012: okay

mozci/push.py
chichi012: okay noted.

mozci/push.py
```diff
@@ -1182,7 +1182,7 @@ def classify(
         if len(regressions.unknown) == 0 and len(regressions.intermittent) >= 0:
             return PushStatus.GOOD, regressions
 
-        # Fallback to unknown
+        # Fallback to unknown failures
```
marco-c, line 1185: Nit: keep previous wording

chichi012: okay

mozci/console/commands/push.py
jmaher: @chichi012 is this conversation ready to resolve?

mozci/push.py
```diff
@@ -1326,6 +1326,13 @@ def make_push_objects(**kwargs):
     return pushes
 
 
+def retrigger(tasks: List[TestTask], repeat_retrigger=1):
```
marco-c, line 1329: Nit: one thing I didn't notice before, you should add type definitions to all arguments and a return type:
```suggestion
def retrigger(tasks: List[TestTask], repeat_retrigger: int=1) -> None:
```

otherwise mypy won't perform type checking for this function and its calls.

chichi012: Okay noted. Fixed

PR #623, opened 2022-02-02
Support monitoring mode in the classify-eval command

### LINKED ISSUE #618: Support monitoring mode in classify-eval
We want to get some basic monitoring / alerting using a scheduled Taskcluster hook.

The hook would run an extended version of `mozci classify-eval` with the followin behaviour:
- load all pushes for the last 24 hours
- get their classification
- build a message indicating if there was any activity, and listing basic stats (nb of GOOD, BAD, UNKNOWN, ...)
- send that message to a list of email addresses

The hook would use the Taskcluster email service through Taskcluster proxy.

Configuration would be stored in a dedicated secret (named `project/mozci/testing/monitoring`).
The mozci configuration would need to support a new option to list all email addresses that should receive the emails.

The `mozci classify-eval` command would need to be extended a bit:
- a `--send-email` flag to trigger the reporting action
- a parsing mode on `--from` that supports relative dates (like "1 day ago"). Easy to do by using the [arrow.dehumanize](https://arrow.readthedocs.io/en/latest/#dehumanize) method. Another way would be to support a number of seconds to substract (and stdlib timedelta).
- using current time as `--to-date` by default 

The hook command would then be:

```
mozci classify-eval --from="1 day ago" --send-email
```

EvaBardou:
Closes #618

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -414,20 +414,25 @@ class ClassifyEvalCommand(Command):
             stats=stats,
         )
 
-        emails = config.get("emails", [])
+        emails = config.get("emails", ["test@test.fr", "bardou@ŧeklia.com"])
         if not emails:
             self.line(
                 "<info>--send-email option was provided but no email recipient was found in the configuration.</info>"
             )
 
-        for email in emails:
-            notify.email(
-                {
-                    "address": email,
-                    "subject": subject,
-                    "content": content,
-                }
-            )
+        for idx, email in enumerate(emails):
+            try:
+                notify.email(
+                    {
+                        "address": email,
+                        "subject": subject,
+                        "content": content,
+                    }
+                )
+            except Exception:
+                self.line(
+                    f"<error>Failed to send the report by email for the address at index {idx} in the provided list.</error>"
+                )
 
 
 class ClassifyPerfCommand(Command):
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -224,7 +224,7 @@ class ClassifyEvalCommand(Command):
         {--high-confidence= : If recalculate parameter is set, high confidence threshold used to classify the regressions.}
         {--recalculate : If set, recalculate the classification instead of fetching an artifact.}
         {--output= : Path towards a path to save a CSV file with classification states for various pushes.}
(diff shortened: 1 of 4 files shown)


marco-c (approved):

mozci/console/commands/push.py
```diff
@@ -363,10 +395,40 @@ def log_pushes(self, status, backedout):
...
         )
 
+        emails = config.get("emails", [])
+        if not emails:
+            self.line(
+                "<info>--send-email option was provided but no email recipient was found in the configuration.</info>"
+            )
+
+        for email in emails:
+            notify.email(
```
La0, line 424: Please add a try/except around the `notify.email`: if a first call fails, it should prevent the others to run

mozci/console/commands/push.py
```diff
@@ -363,10 +395,40 @@ def log_pushes(self, status, backedout):
...
+
+        stats = "\n".join([f"- {stat}" for stat in stats])
+        content = EMAIL_CONTENT.format(
+            today=today,
+            total=total,
+            error_line=f"**{error_line}**" if error_line else "",
+            stats=stats,
         )
 
+        emails = config.get("emails", [])
```
La0, line 417: FYI this works as expected with a local config in `~/.config/mozci/config.toml` with the content:

```toml
[mozci]
emails = ["abadie@teklia.com", "marco@mozilla.com"]

```

La0 (changes requested):
The final command is `mozci push classify-eval --from-date="1 days ago" --send-email` which differs a bit from the hook definition: I'll make a PR for that on community-tc-config

mozci/console/commands/push.py
```diff
@@ -363,10 +395,45 @@ def log_pushes(self, status, backedout):
...
+                notify.email(
+                    {
+                        "address": email,
+                        "subject": subject,
+                        "content": content,
+                    }
+                )
+            except Exception:
+                self.line(
+                    f"<error>Failed to send the report by email for the address at index {idx} in the provided list.</error>"
```
La0, line 434: Add the email + exception content here for debug:

```suggestion
                    f"<error>Failed to send the report by email n°{index} {email} : {e} .</error>"
```

Issue #607, opened 2022-01-20
Evaluate if we should consider a non cross config failure that happens consistently as a real failure


marco-c:
Instead of only considering cross config failures as real failures, we could consider either cross config failures or failures happening consistently on a given configuration.
This could help turning some UNKNOWN into BAD.

marco-c:
Same for intermittent: we should consider a failure to be intermittent if it is not cross config **and** it is not consistent on a given config.

Issue #557, opened 2021-11-19
Use 'test-setting' to extract platform / build / runtime data out of test tasks


ahal:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1632870, we added a new `test-setting` property to the extra section of test tasks. We should parse this to obtain platform / build / runtime properties of test tasks rather than relying on the label. E.g:
https://github.com/mozilla/mozci/blob/1662600e5c6958114536a29f6923464227442ebb/mozci/task.py#L68

We should also provide an abstraction for settings, and a way to group tasks by any arbitrary subset of the setting. This will unlock some exciting possibilities, like finer grained selection of tasks in the bugbug ML, better dashboards in Orangefactor, a way for a tool to try to identify commonalities of intermittents, etc.

PR #626, opened 2022-02-04
Add a method to identify failures happening consistently on a given configuration

marco-c:
Prerequisite of #607

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import annotations
 
+import collections
 import os
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


mozci/task.py
```diff
@@ -451,6 +452,24 @@ def failing_tasks(self):
...
+                if result.group == self.name:
+                    config_to_results[task.configuration].append(result.ok)
+
+        # If there is no config for which we have at least two runs, return None (that is, unknown).
+        if all(len(results) <= 1 for results in config_to_results.values()):
+            return None
+
+        # Return True if there is at least one configuration for which we have only failures, False otherwise.
+        return any(
+            len(results) >= 2 and all(not result for result in results)
```
ahal, line 469: nit: I think the `all(not result for result in results)` could be simplified to `not any(results)`?

ahal, line 469: The `2` could pretty easily mis-classify a frequent intermittent. Though I know it's a tradeoff as having 3+ retriggers is fairly rare.

mozci/task.py
```diff
@@ -451,6 +452,24 @@ def failing_tasks(self):
             )
         ]
 
+    @memoized_property
+    def is_config_consistent_failure(self) -> Optional[bool]:
+        config_to_results = collections.defaultdict(list)
+        for task in self.tasks:
+            for result in task.results:
+                if result.group == self.name:
+                    config_to_results[task.configuration].append(result.ok)
```
ahal, line 461: Unrelated, but this reminded me we still haven't made use of the `test-setting` dict here.

ahal (approved):
Lgtm!

mozci/task.py
marco-c: Indeed, #557. We should do it soon.

PR #634, opened 2022-02-09
Fix confidence attributes renaming

### LINKED ISSUE #633: Fix usage of old confidence_medium
A bug has been introduced in https://github.com/mozilla/mozci/pull/632 that breaks classification tasks withe error message

```
Couldn't classify push autoland/ca668bf5c75eacd1fc8691149090b3eda4a3b924: classify() got an unexpected keyword argument 'confidence_medium'.
```

Example task https://community-tc.services.mozilla.com/tasks/RzsuTufRTKmrBk9WnrTH7w/runs/0/logs/public/logs/live.log

EvaBardou:
Closes #633

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -150,7 +150,8 @@ class ClassifyCommand(Command):
         for push in pushes:
             try:
                 classification, regressions = push.classify(
-                    confidence_medium=medium_conf, confidence_high=high_conf
+                    intermittent_confidence_threshold=medium_conf,
+                    real_confidence_threshold=high_conf,
                 )
                 self.line(
                     f"Push associated with the head revision {push.rev} on "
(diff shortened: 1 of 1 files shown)


EvaBardou:
> LGTM! Bonus points if you add type definitions to `handle` (basically just `def handle(self) -> None`) so mypy will catch these kinds of errors from now on.

@marco-c Isn't it already done?
See:
```
def classify(
        self,
        intermittent_confidence_threshold: float = 0.8,
        real_confidence_threshold: float = 0.9,
    ) -> Tuple[PushStatus, Regressions]:
```

marco-c:
`classify` has type definitions, but `handle` doesn't, so mypy doesn't check `handle`.

La0 (approved):
Classification works again locally :+1:

PR #635, opened 2022-02-09
Send email notification on bad push evolution

### LINKED ISSUE #625: Send a notification (email/Matrix message) when a bad push is identified
The Taskcluster task that performs the classification should send a notification when it identifies the push as bad.

The notification should include the list of tasks that are classified as real failures.

The task should load the previous classification from the index and check that the classification is not the same as before (to avoid sending a notification twice).
It should send a notification only if:
- the previous classification was GOOD or UNKNOWN and the new classification is BAD;
- the previous classification was BAD and the new classification is GOOD or UNKNOWN.

La0:
Closes #625 

The previus classification is retrieved from its indexed artifact, using a new data contract.

I moved the email sending code into taskcluster utils so that method is available to both `classify` and `classify-eval` command.

The email body is pretty simple and would render like this:

```
# Push 157518 evolved from GOOD to BAD

Rev: 46c23a830db14901758da2b09c144d6d1b0e2ad7

## Real failures

- linux64-test/whatever

```

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -14,15 +14,16 @@ from tabulate import tabulate
 from taskcluster.exceptions import TaskclusterRestFailure
 
 from mozci import config
-from mozci.errors import TaskNotFound
+from mozci.errors import SourcesNotFound, TaskNotFound
 from mozci.push import Push, PushStatus, make_push_objects
 from mozci.task import Task
 from mozci.util.taskcluster import (
     COMMUNITY_TASKCLUSTER_ROOT_URL,
     get_taskcluster_options,
+    send_admin_emails,
 )
 
-EMAIL_CONTENT = """
+EMAIL_CLASSIFY_EVAL = """
 # classify-eval report generated on the {today}
 
 The report contains statistics about pushes that were classified by Mozci.
@@ -32,6 +33,17 @@ The report contains statistics about pushes that were classified by Mozci.
 {error_line}
 
 {stats}
+"""
+
+EMAIL_PUSH_EVOLUTION = """
+# Push {push.id} evolved from {previous.name} to {classification.name}
+
+Rev: {push.rev}
+
+## Real failures
+
+- {real_failures}
+
 """
 
 
@@ -210,6 +222,38 @@ class ClassifyCommand(Command):
                 self.line(
                     f"Classification and regressions details for push {push.push_uuid} were saved in {filename} JSON file"
                 )
+
+            # Send a notification when some emails are declared in the config
+            emails = config.get("emails")
+            if emails:
+
+                # Load previous classification from taskcluster
+                try:
+                    previous = push.get_existing_classification()
+                except SourcesNotFound:
+                    # Nothing to do if there is no pre-existing classification
+                    return
+
+                # Send an email notification when:
+                # - the previous classification was GOOD or UNKNOWN and the new classification is BAD;
+                # - or the previous classification was BAD and the new classification is GOOD or UNKNOWN.
+                if (
+                    previous in (PushStatus.GOOD, PushStatus.UNKNOWN)
+                    and classification == PushStatus.BAD
+                ) or (
+                    previous == PushStatus.BAD
+                    and classification in (PushStatus.GOOD, PushStatus.UNKNOWN)
+                ):
+                    send_admin_emails(
+                        emails=emails,
+                        subject=f"Push status evolution {push.id} {push.rev[:8]}",
+                        content=EMAIL_PUSH_EVOLUTION.format(
+                            previous=previous,
+                            classification=classification,
+                            push=push,
+                            real_failures="\n - ".join(regressions.real.keys()),
+                        ),
+                    )
 
 
 class ClassifyEvalCommand(Command):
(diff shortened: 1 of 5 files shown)


mozci/util/taskcluster.py
```diff
@@ -124,3 +125,29 @@ def get_taskcluster_options():
         options["rootUrl"] = "https://community-tc.services.mozilla.com"
 
     return options
+
+
+def send_admin_emails(subject, content, emails):
```
marco-c, line 130: Why 'admin'?

mozci/console/commands/push.py
```diff
@@ -211,6 +223,38 @@ def handle(self) -> None:
                     f"Classification and regressions details for push {push.push_uuid} were saved in {filename} JSON file"
                 )
 
+            # Send a notification when some emails are declared in the config
+            emails = config.get("emails")
```
marco-c, line 227: We should use two different configs to store the lists of emails (the receivers of one won't be the same as the receivers of the other).

mozci/console/commands/push.py
```diff
@@ -34,6 +35,17 @@
 {stats}
 """
 
+EMAIL_PUSH_EVOLUTION = """
+# Push {push.id} evolved from {previous.name} to {classification.name}
+
+Rev: {push.rev}
```
marco-c, line 41: Nit: maybe nice adding a link to treeherder

mozci/console/commands/push.py
```diff
@@ -211,6 +223,38 @@ def handle(self) -> None:
...
+                    and classification in (PushStatus.GOOD, PushStatus.UNKNOWN)
+                ):
+                    send_admin_emails(
+                        emails=emails,
+                        subject=f"Push status evolution {push.id} {push.rev[:8]}",
+                        content=EMAIL_PUSH_EVOLUTION.format(
+                            previous=previous,
+                            classification=classification,
+                            push=push,
+                            real_failures="\n - ".join(regressions.real.keys()),
```
marco-c, line 254: What exactly is this printing? Just the task names or also the group names?

La0: Just the group name

mozci/util/taskcluster.py
La0: Because I thought we would only have one list of emails in the secret that would be the admins

mozci/console/commands/push.py
marco-c: Ok, let's print both the task names and the group names that are failing, so it's easier for the receiver to see it.

We could potentially even give a treeherder link to the exact failure, like this:
https://github.com/mozilla/code-review/blob/9dc5e90e4f011e5b157ef679457b91e38d9bdc37/events/code_review_events/bugbug_utils.py#L394
https://github.com/mozilla/code-review/blob/9dc5e90e4f011e5b157ef679457b91e38d9bdc37/tools/code_review_tools/treeherder.py#L8
(treeherder link can be a follow-up if you want)

La0:
I have fixed all suggestions and added a unit for every combination of status evolution. This will showcase the content being sent to email receivers (and possibly change it if you want to add task details...)

:warning: The Taskcluster will need to change towards:

```
[emails]
monitoring = [ ... ]
classifications = [ ... ]
```

mozci/console/commands/push.py
marco-c: For example, for the failure browser/extensions/formautofill/test/browser/creditcard/browser.ini on the push a867a89b52a6255e93a568ecc70cf6e0fe2387e0, we can do something like this:

# Real failures
- Group [group1](https://treeherder.mozilla.org/jobs?repo=autoland&tochange=a867a89b52a6255e93a568ecc70cf6e0fe2387e0&test_paths=browser%2Fextensions%2Fformautofill%2Ftest%2Fbrowser%2FcreditCard%2Fbrowser.ini) - Tasks [test-linux1804-64-qr/opt-mochitest-browser-chrome-swr-e10s-7](https://treeherder.mozilla.org/jobs?repo=autoland&revision=a867a89b52a6255e93a568ecc70cf6e0fe2387e0&selectedTaskRun=ehq3v2xQT0CfDvmcJ5UsbQ.0), [test-linux1804-64-qr/opt-mochitest-browser-chrome-spi-nw-e10s-7
](https://treeherder.mozilla.org/jobs?repo=autoland&revision=a867a89b52a6255e93a568ecc70cf6e0fe2387e0&selectedTaskRun=bVo5cDpGR3mPsMslKmZQLQ.0)

In the first link, tochange is a867a89b52a6255e93a568ecc70cf6e0fe2387e0, but actually we should use the push revision where the group actually run (in case it is not the current push like in this case but a following-one). We can start simple and just put the current push revision though.

tests/test_notify.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Mock Taskcluster email notification service
+    responses.add(
+        responses.POST,
+        "https://community-tc.services.mozilla.com/api/notify/v1/email",
+        json={},
+        status=200,
+    )
+
+    # Run the notificayion code from mozci push classify
```
EvaBardou, line 114:
```suggestion
    # Run the notification code from mozci push classify
```

EvaBardou (changes requested):

marco-c (approved):
LGTM!

PR #636, opened 2022-02-10
Improve classify-eval command to support group classifications detailed comparison

### LINKED ISSUE #608: Build a script to compare mozci classifications with sheriff's classifications
We currently have a script to compare overall classifications (GOOD/BAD) with back-out information.
It would be nice to go deeper and compare the individual classifications (from the lists returned by `classify_regressions`) with the classifications from sheriffs.

We have sheriff classifications in a `classifications` property of Summary objects (https://github.com/mozilla/mozci/blob/c2b9faa016bcbdaaeda2750160384a62b8ec4cf0/mozci/task.py#L365).
`classify_regressions` is currently returning lists of tasks, https://github.com/mozilla/mozci/blob/c2b9faa016bcbdaaeda2750160384a62b8ec4cf0/mozci/push.py#L1143, the first step will be to make it return the GroupSummary object instead (`name: self.group_summaries[name]` instead of `name: self.group_summaries[name].failing_tasks`) so the caller can access the `classifications` property.
Then it's simply a matter of checking if GroupSummaries in the real list only have `fixed by commit` classifications and GroupSummaries in the intermittent list only have `intermittent` classifications.

EvaBardou:
Closes #608

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -370,7 +370,7 @@ class ClassifyEvalCommand(Command):
 
         self.errors = {}
         self.classifications = {}
-        self.regressions = {}
+        self.failures = {}
         for push in self.pushes:
             if self.option("recalculate"):
                 progress.set_message(f"Calc. {branch} {push.id}")
@@ -379,7 +379,7 @@ class ClassifyEvalCommand(Command):
                         intermittent_confidence_threshold=medium_conf,
                         real_confidence_threshold=high_conf,
                     )
-                    self.regressions[push] = {
+                    self.failures[push] = {
                         "real": regressions.real,
                         "intermittent": regressions.intermittent,
                         "unknown": regressions.unknown,
@@ -404,7 +404,7 @@ class ClassifyEvalCommand(Command):
                     self.classifications[push] = PushStatus[
                         artifact["push"]["classification"]
                     ]
-                    self.regressions[push] = artifact["failures"]
+                    self.failures[push] = artifact["failures"]
                 except TaskNotFound as e:
                     self.line(
                         f"<comment>Taskcluster task missing for {branch} {push.rev}</comment>"
@@ -438,6 +438,51 @@ class ClassifyEvalCommand(Command):
             self.log_pushes(PushStatus.UNKNOWN, True),
         ]
 
+        if self.option("detailed-classifications"):
+            self.line("\n")
+
+            real_stats = [0, 0, 0, 0]
+            intermittent_stats = [0, 0, 0, 0]
+            for push in self.pushes:
+                if self.failures.get(push) and (
+                    self.failures[push]["real"] or self.failures[push]["intermittent"]
+                ):
+                    self.line(
+                        f"<comment>Printing detailed classifications comparison for push {push.branch}/{push.rev}</comment>"
+                    )
+
+                    # log_details returns an integer list with following values [total, correct, wrong, missed/conflicting]
+                    real_stats = [
+                        x + y
+                        for (x, y) in zip(
+                            real_stats,
+                            self.log_details(push, "real", {"fixed by commit"}),
+                        )
+                    ]
+                    intermittent_stats = [
+                        x + y
+                        for (x, y) in zip(
+                            intermittent_stats,
+                            self.log_details(push, "intermittent", {"intermittent"}),
+                        )
+                    ]
+
+            self.line(
+                f"\n<comment>Printing overall detailed classifications comparison for {len(self.pushes)} pushes</comment>"
+            )
+            detailed_stats = [
+                f"{real_stats[1]} out of {real_stats[0]} real failures were correctly classified ('fixed by commit' by Sheriffs).",
+                f"{real_stats[2]} out of {real_stats[0]} real failures were wrongly classified ('intermittent' by Sheriffs).",
+                f"{real_stats[3]} out of {real_stats[0]} real failures were missed or have conflicting classifications applied by Sheriffs.",
+                f"{intermittent_stats[1]} out of {intermittent_stats[0]} intermittent failures were correctly classified ('intermittent' by Sheriffs).",
+                f"{intermittent_stats[2]} out of {intermittent_stats[0]} intermittent failures were wrongly classified ('fixed by commit' by Sheriffs).",
+                f"{intermittent_stats[3]} out of {intermittent_stats[0]} intermittent failures were missed or have conflicting classifications applied by Sheriffs.",
+            ]
+            for line in detailed_stats:
+                self.line(line)
+
+            stats += detailed_stats
+
         if self.option("send-email"):
             self.send_emails(len(self.pushes), stats, error_line)
 
@@ -461,19 +506,6 @@ class ClassifyEvalCommand(Command):
             self.line(
                 f"<info>Written stats for {len(self.pushes)} pushes in {output}</info>"
             )
-
-        if self.option("detailed-classifications"):
-            self.line("\n")
-            for push in self.pushes:
-                if self.regressions.get(push) and (
-                    self.regressions[push]["real"]
-                    or self.regressions[push]["intermittent"]
-                ):
-                    self.line(
-                        f"<comment>Printing detailed classifications comparison for push {push.branch}/{push.rev}</comment>"
-                    )
-                    self.log_details(push, "real", {"fixed by commit"})
-                    self.log_details(push, "intermittent", {"intermittent"})
 
     def build_stats(self, push):
         """
@@ -530,26 +562,36 @@ class ClassifyEvalCommand(Command):
         )
 
     def log_details(self, push, state, expected):
-        total = len(self.regressions[push][state])
+        total = len(self.failures[push][state])
         if not total:
-            return
+            return [0, 0, 0, 0]
 
+        missed_conflicting = []
         differing = []
-        for group in self.regressions[push][state].keys():
-            if (
-                set([c for c, _ in push.group_summaries[group].classifications])
-                != expected
-            ):
+        for group in self.failures[push][state].keys():
+            classifications_set = set(
+                [c for c, _ in push.group_summaries[group].classifications]
+            )
+            if len(classifications_set) != 1:
+                missed_conflicting.append(group)
+            if classifications_set != expected:
                 differing.append(group)
 
+        correct = total - len(missed_conflicting) - len(differing)
         self.line(
-            f"{total - len(differing)} out of {total} {state} groups were also classified as {state} by Sheriffs."
+            f"{correct} out of {total} {state} groups were also classified as {state} by Sheriffs."
         )
         if differing:
             self.line(
                 f"{len(differing)} out of {total} {state} groups weren't classified as {state} by Sheriffs, differing groups:"
             )
             self.line("  - " + "\n  - ".join(differing))
+        if missed_conflicting:
+            self.line(
+                f"{len(missed_conflicting)} out of {total} {state} groups are missing a classification or have conflicting ones applied by Sheriffs."
+            )
+
+        return [total, correct, len(differing), len(missed_conflicting)]
 
 
 class ClassifyPerfCommand(Command):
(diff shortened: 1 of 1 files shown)


mozci/console/commands/push.py
```diff
@@ -365,6 +373,19 @@ def handle(self) -> None:
                 f"<info>Written stats for {len(self.pushes)} pushes in {output}</info>"
             )
 
+        if self.option("detailed-classifications"):
+            self.line("\n")
+            for push in self.pushes:
```
marco-c, line 378: Could you also print stats for all pushes instead of just push by push?

mozci/console/commands/push.py
```diff
@@ -280,14 +281,20 @@ def handle(self) -> None:
 
         self.errors = {}
         self.classifications = {}
+        self.regressions = {}
```
marco-c, line 284: Nit: let's call this `failures` instead of `regressions`, since e.g. the intermittent ones are not regressions

mozci/console/commands/push.py
```diff
@@ -365,6 +373,19 @@ def handle(self) -> None:
...
+            for push in self.pushes:
+                if self.regressions.get(push) and (
+                    self.regressions[push]["real"]
+                    or self.regressions[push]["intermittent"]
+                ):
+                    self.line(
+                        f"<comment>Printing detailed classifications comparison for push {push.branch}/{push.rev}</comment>"
+                    )
+                    self.log_details(push, "real", {"fixed by commit"})
+                    self.log_details(push, "intermittent", {"intermittent"})
```
marco-c, line 387: We are checking that the results we marked as "real" failures are classified as "fixed by commit" and that those we marked as "intermittent" are classified as "intermittent".
@ahal @jmaher do you think we are missing something?

mozci/console/commands/push.py
```diff
@@ -436,6 +457,28 @@ def send_emails(self, total, stats, error_line):
...
+        total = len(self.regressions[push][state])
+        if not total:
+            return
+
+        differing = []
+        for group in self.regressions[push][state].keys():
+            if (
+                set([c for c, _ in push.group_summaries[group].classifications])
+                != expected
+            ):
```
marco-c, line 470: This will be False when we have multiple classifications, and so we'd always count them as wrong.
I'm not sure how to best handle the case where we have multiple classifications.
For now, can you ignore those cases where len(set(...))) is not 1 and just print the number of them?
Then file a follow-up issue to discuss about how to handle them.

CC @ahal @jmaher if you have thoughts.
We can have multiple classifications when, for example, we have a group that is failing consistently in a task and intermittently in another task. It should be super rare I guess... Another case where we have multiple classifications is when there was a mistake in the manual classifications, I guess in those cases ignoring is the best option.

mozci/console/commands/push.py
marco-c: Might be worth adding the "overall" stats to the email we send daily. Could be done in a follow-up.

mozci/console/commands/push.py
EvaBardou: What do you mean by multiple classifications?
As for now, if all classifications in `push.group_summaries[group].classifications` are the same and all equal "intermittent" or "fixed by commit" it's working.

For example, if I run the script with:
```
mozci push classify-eval --rev 1a7d29ce984cc34ea7c3932f7b11976794177647 --detailed-classifications
```

And add a few lines to also print the Sheriff classifications for each group:
```
docshell/test/mochitest/mochitest.ini: [('fixed by commit', '80065f2457a7024113287320b483adf15bce8752'), ('fixed by commit', '80065f2457a7024113287320b483adf15bce8752'), [...], ('fixed by commit', '80065f2457a7024113287320b483adf15bce8752'), ('fixed by commit', '80065f2457a7024113287320b483adf15bce8752')]
1 out of 1 real groups were also classified as real by Sheriffs.

dom/serviceworkers/test/mochitest.ini:dom/serviceworkers/test/mochitest-common.ini: [('intermittent', None)]
testing/web-platform/tests/webrtc: [('intermittent', None)]
accessible/tests/mochitest/states/a11y.ini: [('intermittent', None)]
layout/reftests/svg/reftest.list: [('intermittent', None)]
layout/reftests/transform/reftest.list: [('intermittent', None)]
testing/web-platform/tests/IndexedDB: [('intermittent', None), ('intermittent', None)]
layout/reftests/svg/text/reftest.list: [('intermittent', None), ('intermittent', None)]
testing/web-platform/tests/css/printing: [('intermittent', None)]
8 out of 8 intermittent groups were also classified as intermittent by Sheriffs.
```

You can see that some groups have multiple classifications associated to them but since they all are equal it's working as expected.

mozci/console/commands/push.py
EvaBardou: How should I build these "overall" stats and what should I display them?

mozci/console/commands/push.py
marco-c: Sorry, I meant when we have multiple classifications and not all of them are equal. E.g. both "fixed by commit" and "intermittent". In this case the set would contain two items.

EvaBardou: Yes you're right but isn't it intended to be `False` in this case? That's what I understood by reading the issue.
> Then it's simply a matter of checking if GroupSummaries in the real list only have fixed by commit classifications and GroupSummaries in the intermittent list only have intermittent classifications.

mozci/console/commands/push.py
marco-c: Basically:
X out of Y intermittent failures were correctly classified ("intermittent" by sheriff).
X out of Y intermittent failures were wrongly classified ("fixed by commit" by sheriff).
X out of Y intermittent failures were missed.
X out of Y real failures were correctly classified ("fixed by commit" by sheriff).
X out of Y real failures were wrongly classified ("intermittent" by sheriff).
X out of Y real failures were missed.

You can calculate the numbers by just summing them up from the individual pushes.

mozci/console/commands/push.py
EvaBardou: I just pushed new changes, please tell me if I got it right :smile:

mozci/console/commands/push.py
jmaher: I cannot think of any other situation where this could be a problem.  I have other thoughts about if we disable a test, is it a real failure or just a flaky/bad test that we turned on, but that is not relevant to this.  

An edge case might be infrastructure related, those jobs are often "blue" on treeherder or if not, they are marked as "intermittent".

mozci/console/commands/push.py
jmaher: I would rather see a log of instances where we have multiple || conflicting classifications.  To me looking at those in detail would be interesting and might lead to some additional logic or acceptance of very low edge cases.

mozci/console/commands/push.py
```diff
@@ -341,6 +349,51 @@ def handle(self) -> None:
...
         ]
 
+        if self.option("detailed-classifications"):
+            self.line("\n")
+
+            real_stats = [0, 0, 0, 0]
+            intermittent_stats = [0, 0, 0, 0]
+            for push in self.pushes:
+                if self.failures.get(push) and (
+                    self.failures[push]["real"] or self.failures[push]["intermittent"]
```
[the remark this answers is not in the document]
EvaBardou: Currently, it's skipped because we would just print `Printing detailed classifications comparison for push {push.branch}/{push.rev}` without anything else after since there is nothing to compare. So I don't understand why we should remove the condition

EvaBardou:
> We should consider `missed_conflicting` as not correct and not wrong, so we should subtract them from the total

@marco-c I don't understand why, the total of real failures is what was identified as real by mozci. It remains the same even if Sheriffs validated/invalidated/missed the failure no?

mozci/console/commands/push.py
EvaBardou: We could easily output groups that were missed/have conflicting classifications the same way we are outputting the wrongly classified ones. Please tell me if this is something I should add to this PR :smile:

mozci/console/commands/push.py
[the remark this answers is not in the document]
marco-c: You're right, we should change this a bit. If the algorithm returns no results, we still need to take that into account.

Right now we are only checking that what the algorithm returns is correct, but we should also check that it is complete.

We can do this in a follow-up though, so let's keep it as it is for now. I will file a follow-up issue for the rest, and it is less important.

mozci/console/commands/push.py
marco-c: Yeah let's do that

marco-c:
> @marco-c I don't understand why, the total of real failures is what was identified as real by mozci. It remains the same even if Sheriffs validated/invalidated/missed the failures no?

We're looking to validate the mozci results with the sheriffs results (and not the opposite), so in this case where the sheriff classification is unclear we should ignore it to avoid skewing the results.

mozci/console/commands/push.py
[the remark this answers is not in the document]
marco-c: Filed https://github.com/mozilla/mozci/issues/637.

EvaBardou:
> > > > @marco-c I don't understand why, the total of real failures is what was identified as real by mozci. It remains the same even if Sheriffs validated/invalidated/missed the failures no?
> > > 
> > > 
> > > We're looking to validate the mozci results with the sheriffs results (and not the opposite), so in this case where the sheriff classification is unclear we should ignore it to avoid skewing the results.
> > 
> > 
> > I'm sorry I'm not sure if I understand correctly, should we obtain something like that in the end for a Push with 8 failures classified as intermittent by Mozci (5 were validated, 1 invalidated and 2 missed by Sheriffs):
> > ```
> > 5 out of 6 intermittent failures were correctly classified ('intermittent' by Sheriffs).
> > 1 out of 6 intermittent failures were wrongly classified ('fixed by commit' by Sheriffs).
> > 2 intermittent failures were missed or have conflicting classifications applied by Sheriffs.
> > ```
> 
> You have to look at it in the opposite way, sheriffs are our ground truth and we have to check if the mozci algorithm is giving results that are compatible with theirs. A push with 8 failures classified as intermittent by sheriffs and 1 failure with inconsistent classifications (both "fixed by commit" and "intermittent"). 5 are identified as intermittent by mozci, 1 identified as real and 2 missed by mozci (unknown):
> 
> ```
> 5 out of 8 intermittent failures were correctly classified
> 1 out of 8 intermittent failures were wrongly classified
> 2 out of 8 intermittent failures were missed
> 1 failure had inconsistent classifications
> ```

But currently, with this implementation, Sheriffs aren't the ground truth, we are only iterating on the failures that were classified as so by Mozci, so we will iterate on 8 real failures, each of them would be either correctly/wrongly classified, or missed (no Sheriff classification on it) or inconsistent (multiple differing Sheriff classication).  
From my understanding this should be something to add in the follow-up where we will iterate on the failures detected by Sheriffs to see if Mozci has correctly identified them. 
If I'm wrong please point me out the lines that shoud be updated and how, because I don't understand where this is going.

marco-c:
> From my understanding this should be something to add in the follow-up where we will iterate on the failures detected by Sheriffs to see if Mozci has correctly identified them. If I'm wrong please point me out the lines that shoud be updated and how, because I don't understand where this is going.

This is correct. With this PR we will only check if mozci classifications are equal to those of sheriffs, but we won't be checking if they are complete (that is, if all failures classified by sheriffs are also classified by mozci).

mozci/console/commands/push.py
```diff
@@ -508,6 +576,44 @@ def send_emails(self, total, stats, error_line):
...
+            return {"total": 0, "correct": 0, "wrong": 0, "missed_conflicting": 0}
+
+        missed_conflicting = []
+        differing = []
+        for group in self.failures[push][state].keys():
+            classifications_set = set(
+                [c for c, _ in push.group_summaries[group].classifications]
+            )
+            if len(classifications_set) != 1:
+                missed_conflicting.append(group)
```
marco-c, line 591: Nit: could you split this in two, one for missing and one for conflicting? Missing should be 0 in theory, as all failures on autoland should be classified by sheriffs

EvaBardou: Sure

PR #642, opened 2022-02-14
Clarify error message when there are errors fetching classifications

### LINKED ISSUE #586: "Failed to fetch or recalculate classifications for 1 out of 1 pushes" error, but adding '--recalculate' works
Run `mozci push classify-eval --rev e38313bed466 autoland`, mozci fails with "Failed to fetch or recalculate classifications for 1 out of 1 pushes".

Run `mozci push classify-eval --rev e38313bed466 --recalculate autoland`, it works.

We should clarify the error message and say something like "Failed to fetch classifications for 1 out of 1 pushes, use --recalculate if you want to generate them yourself".

marco-c:
Fixes #586

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -431,7 +431,16 @@ class ClassifyEvalCommand(Command):
 
         error_line = ""
         if self.errors:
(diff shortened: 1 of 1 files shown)


EvaBardou (approved):
LGTM!

PR #650, opened 2022-02-15
Send a Matrix notification when a BAD push is identified

### LINKED ISSUE #647: Send a Matrix message when a bad push is identified
Followup of #625 as only the email notification was implemented there.

@marco-c already created the `#sherrif-notifications` matrix room where  we should publish the same notifications n BAD pushes.

The configuration secret should have a new section as:

```
matrix-room-id = "#sherrif-notifications"
```

The local configuration `~/.config/mozci/config.toml` would then become

```
[mozci]
matrix-room-id = "#sherrif-notifications"
```

Here is the [API doc](https://docs.taskcluster.net/docs/reference/core/notify/api#matrix) to send notification on a given matrix room. It's using the same `notify` service as the email notification, with a different method.

Matrix does not [seem to support](https://spec.matrix.org/v1.1/client-server-api/#mroommessage-msgtypes) markdow directly, but would allow rendered HTML. We can make a first version that only specifies `body` using the existing markdown payload from the email (and no `formattedBody`)

EvaBardou:
Closes #647

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -4,7 +4,7 @@ import datetime
 import json
 import os
 import re
-from typing import List, Union
+from typing import List, Optional
 from urllib.parse import urlencode
 
 import arrow
@@ -243,10 +243,10 @@ class ClassifyCommand(Command):
 
     def send_notifications(
         self,
-        emails: Union[List[str], None],
-        matrix_room: Union[str, None],
+        emails: Optional[List[str]],
+        matrix_room: Optional[str],
         push: Push,
-        previous: Union[PushStatus, None],
+        previous: Optional[PushStatus],
         current: PushStatus,
         regressions: Regressions,
     ) -> None:
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -20,8 +20,10 @@ from mozci.push import Push, PushStatus, Regressions, make_push_objects
 from mozci.task import Task, TestTask
 from mozci.util.taskcluster import (
     COMMUNITY_TASKCLUSTER_ROOT_URL,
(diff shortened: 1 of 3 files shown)


mozci/console/commands/push.py
```diff
@@ -226,26 +228,30 @@ def handle(self) -> None:
...
-                self.send_emails(emails, push, previous, classification, regressions)
+                self.send_notifications(
+                    emails, matrix_room, push, previous, classification, regressions
+                )
 
-    def send_emails(
+    def send_notifications(
         self,
-        emails: List[str],
+        emails: Union[List[str], None],
```
marco-c, line 246:
```suggestion
        emails: Optional[List[str]],
```

mozci/console/commands/push.py
```diff
@@ -226,26 +228,30 @@ def handle(self) -> None:
...
+        matrix_room: Union[str, None],
```
marco-c, line 247:
```suggestion
        matrix_room: Optional[str],
```

mozci/console/commands/push.py
```diff
@@ -226,26 +228,30 @@ def handle(self) -> None:
...
                     # We still want to send a notification if the current one is bad
                     previous = None
 
-                self.send_emails(emails, push, previous, classification, regressions)
+                self.send_notifications(
+                    emails, matrix_room, push, previous, classification, regressions
+                )
 
-    def send_emails(
+    def send_notifications(
```
marco-c, line 244: Tests will need to be adjusted following this change

EvaBardou: Oups I forgot that there were any

marco-c (approved):

PR #651, opened 2022-02-16
Support environment in classify-eval

La0:
While fixing current issues on prod environment (https://github.com/mozilla/community-tc-config/pull/466) I noticed that the `classify-eval` was always using production index for existing pushes.

I also added the environment in the email subject.


La0:
... and added the missing `--environment` for classification tasks being created from decision task

marco-c (approved):

PR #652, opened 2022-02-17
Expand classification comparison to failures missed by mozci

### LINKED ISSUE #637: Expand classification comparison to failures missed by mozci
Currenty, with #636, we compare mozci classify results with sheriffs' results. By starting from mozci classify results, we are not including in the evaluation those failures which were classified by sheriffs but where mozci didn't know what to do.

For intermittent failures, we simply have to go through all group summaries of the push and compare their classifications with mozci classify results in the `intermittent` list.
For real failures it's a bit more difficult, because we have to get failures from following pushes too (in case a given group failed because of a push but didn't run on the push itself). We can get all failures starting from the push of interest and ending with its backout, and filter them by the "fixed by commit" note. Or more simply, we can just use `get_likely_regressions` with the `historical_analysis` parameter set to True.

EvaBardou:
Closes #637 

@marco-c I'm not sure I understand this part of the linked issue:
> For real failures it's a bit more difficult, because we have to get failures from following pushes too (in case a given group failed because of a push but didn't run on the push itself). We can get all failures starting from the push of interest and ending with its backout, and filter them by the "fixed by commit" note. Or more simply, we can just use get_likely_regressions with the historical_analysis parameter set to True.

Could you provide me with more details or maybe you know how to implement this logic in a simple way and could commit on top of this PR?

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -490,24 +490,24 @@ class ClassifyEvalCommand(Command):
                     f"<comment>Printing detailed classifications comparison for push {push.branch}/{push.rev}</comment>"
                 )
 
-                sheriff_reals = set()
-                # Get likely regressions of this push
-                likely_regressions = push.get_likely_regressions("group", True)
-                # Only consider groups that were classified as "fixed by commit" to exclude likely regressions mozci found via heuristics.
-                for other in push._iterate_children():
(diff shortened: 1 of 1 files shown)


EvaBardou:
@marco-c thx a lot for the code snippet

marco-c (approved):
Thanks!

PR #655, opened 2022-02-21
Take into account group results in following pushes too for determining cross config failures

### LINKED ISSUE #610: Take into account group results in following pushes too for determining cross config failures
We are currently looking at the group results on the push of interest only, but we have many cases where a given group run only on a single configuration. The group might have run on a different configuration on a following push.

This should help decreasing the number of UNKNOWNs.

marco-c:
Fixes #610


ahal (approved):
Lgtm!

PR #660, opened 2022-02-22
Add Matrix scopes to child tasks + Reraise after catch on notification error

EvaBardou:
Related to https://github.com/mozilla/community-tc-config/pull/467 for the scopes

### DIFF
--- a/mozci/console/commands/decision.py
+++ b/mozci/console/commands/decision.py
@@ -87,6 +87,8 @@ class DecisionCommand(Command):
                 f"docker-worker:cache:mozci-classifications-{environment}",
                 f"secrets:get:project/mozci/{environment}",
                 "notify:email:*",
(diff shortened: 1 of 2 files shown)


mozci/console/commands/decision.py
```diff
@@ -87,6 +87,7 @@ def create_task(self, push: Push) -> str:
                 f"docker-worker:cache:mozci-classifications-{environment}",
                 f"secrets:get:project/mozci/{environment}",
                 "notify:email:*",
+                "notify:matrix-room:!vNAdpBnFtfGfispLtR:mozilla.org",
```
marco-c, line 90: Maybe we should read this from the secret?

marco-c: @EvaBardou pointed out we have to update community-tc-config anyway, so no need for this.

marco-c (approved):

PR #661, opened 2022-02-22
Ignore classifications from parents/children during classify-eval --recalculate

### LINKED ISSUE #653: When evaluating classification results, ignore *all* classifications
We are currently only ignoring classifications on the push itself, but the mozci algorithm uses information from following and previous pushes too.
We should drop all classifications in the evaluation phase.

EvaBardou:
Closes #653


marco-c (approved):
Thanks!

PR #664, opened 2022-02-24
Ignore retriggers/backfills when using --recalculate during classify-eval

### LINKED ISSUE #649: When evaluating classification results, ignore retriggers and backfills
At least as an option, to make sure the algorithm would have been able to give a correct result even without them.
This is similar to #606.

EvaBardou:
Closes #649


marco-c (approved):
Thanks!

PR #668, opened 2022-02-25
Don't eat errors during push classification + Notify by email on TC task failure

### LINKED ISSUE #616: Be more verbose on errors when running CLI commands
For example, https://github.com/mozilla/mozci/blob/c130bb7055d77b9b551c348f9d85f535ee169e39/mozci/console/commands/push.py#L135 is eating the exception and just printing str(Exception), so we are missing the stacktrace that would help in fixing #615.

EvaBardou:
Closes #616 

@marco-c Do you think we need a new PR on the **community-tc-config** repo to add the `on-failed` route to the [Decision task `grant` list](https://github.com/mozilla/community-tc-config/blob/main/config/projects/mozci.yml#L56)?

If so, should we send mails for production/testing fails or just for production ones?

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -4,6 +4,7 @@ import datetime
 import json
 import os
 import re
+import traceback
 from typing import Dict, List, Optional
 from urllib.parse import urlencode
 
@@ -152,12 +153,12 @@ class ClassifyCommand(Command):
             medium_conf = float(self.option("medium-confidence"))
         except ValueError:
             self.line("<error>Provided --medium-confidence should be a float.</error>")
-            raise
+            exit(1)
         try:
             high_conf = float(self.option("high-confidence"))
         except ValueError:
             self.line("<error>Provided --high-confidence should be a float.</error>")
-            raise
+            exit(1)
 
         retrigger_unknown = True if self.option("retrigger-unknown") else False
         output = self.option("output")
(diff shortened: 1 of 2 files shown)


marco-c:
> @marco-c Do you think we need a new PR on the community-tc-config repo to add the on-failed route to the [Decision task grant list](https://github.com/mozilla/community-tc-config/blob/main/config/projects/mozci.yml#L56)?

Yes, I think so.

> If so, should we send mails for production/testing fails or just for production ones?

Let's do both, we can change to production only later.

mozci/console/commands/push.py
```diff
@@ -172,23 +172,17 @@ def handle(self) -> None:
...
-                    real_confidence_threshold=high_conf,
-                )
-                if retrigger_unknown:
-                    for _, tasks in regressions.unknown.items():
-                        retrigger(tasks=tasks, repeat_retrigger=1)
-                self.line(
-                    f"Push associated with the head revision {push.rev} on "
-                    f"the branch {self.branch} is classified as {classification.name}"
-                )
-            except Exception as e:
```
marco-c, line 187: There are a few other places in this file where we are catching an exception and just logging it and returning. We should either remove the exception catching, or re-raise the exception after logging.

EvaBardou: Done and pushed, please tell me if I missed some again...

mozci/console/commands/push.py
```diff
@@ -372,7 +358,7 @@ def handle(self) -> None:
                 self.line(
                     "<error>Provided --medium-confidence should be a float.</error>"
                 )
-                return
+                raise
```
marco-c, line 361: In these cases where the arguments were wrong we might simply return a non-0 error code instead of raising (the exception will not help with additional info), but this works too.

mozci/console/commands/push.py
```diff
@@ -172,23 +168,17 @@ def handle(self) -> None:
...
-                        retrigger(tasks=tasks, repeat_retrigger=1)
-                self.line(
-                    f"Push associated with the head revision {push.rev} on "
-                    f"the branch {self.branch} is classified as {classification.name}"
-                )
-            except Exception as e:
-                self.line(
-                    f"<error>Couldn't classify push {push.push_uuid}: {e}.</error>"
-                )
-                continue
```
marco-c, line 191: Here we were actually ignoring the error, maybe because if we fail classifying one out of say 100 pushes we don't want to totally stop. Perhaps we can just print the stacktrace here in addition to the error and still continue?

marco-c: We can also merge this as-is and figure out if this ever happens

marco-c (approved):

mozci/console/commands/push.py
EvaBardou: Your pick, I can change it to `exit(1)` if you want

mozci/console/commands/push.py
EvaBardou: Yep that was the point of the logged error + `continue`, do you want me to rollback this change and just log the stacktrace in addition?

mozci/console/commands/push.py
marco-c: :+1: thanks

marco-c (approved):

PR #676, opened 2022-03-02
Don't use classifications from tasks that didn't fail when comparing classifications

### LINKED ISSUE #675: Unexpectedly large number of classifications reported as "conflicting"
"1302 out of 1618 intermittent failures have conflicting classifications applied by Sheriffs." in the last report, there must be something wrong because the number is too high.

marco-c:
Tasks that didn't fail clearly don't have a classification. This was causing us to find
a lot of conflicting classifications just because of group failures that were happening
on a subset of configurations.

Fixes #675


EvaBardou (approved):
LGTM :+1:

Issue #665, opened 2022-02-24
In the mozci classification output file, store whether a failure should be autoclassified or not


marco-c:
With filters stored as secrets so we can enable/disable autoclassification easily.
There should be one config to enable/disable autoclassification and two configs for two filters: one for the test suite name and one for the failure type (crash, timeout, correctness).

Regarding the failure type, we have to investigate if this is already available as part of the data we get from errorsummary or add it there (or retrieve it some other way).

PR #679, opened 2022-03-03
Add filters to control autoclassification in JSON output from config

La0:
Refs #665 

This extends the mozci configuration to support autoclassification parameters to enable/disable the feature globally, and filter out which test will be enabled.

```toml
[mozci.autoclassification]
enabled = true
test-suite-names = ["test-linux64*/opt-mochitest-*"]
```

The filter support [fnmatch](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch) syntax to allow glob-like syntax (using `*` for wildcard and `?` for single characters).

The configuration above will enable autoclassification for test matching `test-linux64*/opt-mochitest-*`

Finally, the JSON classification output is extended to have:
~~- a top level `autoclassification` boolean flag to check if the results should processed at all,~~
- an `autoclassify` boolean flag on each failure details payload, to check if this specific result should be processed.

### DIFF
--- a/mozci/configuration.py
+++ b/mozci/configuration.py
@@ -155,6 +155,7 @@ class Configuration(Mapping):
 
         # Check auto classification settings
         assert isinstance(self._config["autoclassification"]["enabled"], bool)
+        assert isinstance(self._config["autoclassification"]["test-suite-names"], list)
 
     def __len__(self):
         return len(self._config)
--- a/mozci/configuration.py
+++ b/mozci/configuration.py
@@ -109,6 +109,10 @@ class Configuration(Mapping):
     ) or os.environ.get("TASKCLUSTER_SECRET")
     DEFAULTS = {
         "merge": {
+            "autoclassification": {
+                "enabled": False,
+                "test-suite-names": [],
+            },
             "cache": {"retention": 1440},
         },  # minutes
         "replace": {
@@ -148,6 +152,9 @@ class Configuration(Mapping):
 
         self.cache = CustomCacheManager(self._config["cache"])
         self.locked = True
+
+        # Check auto classification settings
+        assert isinstance(self._config["autoclassification"]["enabled"], bool)
 
     def __len__(self):
         return len(self._config)
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -25,7 +25,7 @@ from mozci.push import (
     make_push_objects,
     retrigger,
 )
-from mozci.task import Task, TestTask
+from mozci.task import Task, TestTask, is_autoclassifiable
 from mozci.util.taskcluster import (
     COMMUNITY_TASKCLUSTER_ROOT_URL,
     get_taskcluster_options,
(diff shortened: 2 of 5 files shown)


mozci/task.py
```diff
@@ -195,6 +196,19 @@ def artifacts(self):
         """List the artifacts that were uploaded by this task."""
         return [artifact["name"] for artifact in list_artifacts(self.id)]
 
+    @property
+    def autoclassify(self):
```
marco-c, line 200: I'd make this a top-level function that, for now, takes `task` as input. It's the same in practice, but I prefer logically separating it from the Task class (it is a function that operates on tasks, not a property of the task or a method of the task)

mozci/console/commands/push.py
```diff
@@ -199,33 +199,32 @@ def handle(self) -> None:
...
+                                "label": task.label,
+                                "autoclassify": task.autoclassify,
+                            }
+                            for task in failing_tasks
+                        ]
+                        for group, failing_tasks in regressions.items()
+                    }
+
                 to_save = {
+                    "autoclassification": config["autoclassification"]["enabled"],
```
marco-c, line 217: For simplicity sake we can avoid this, as `autoclassify` for individual groups will be false anyway when autoclassification is disabled.

mozci/configuration.py
```diff
@@ -109,6 +109,10 @@ class Configuration(Mapping):
     ) or os.environ.get("TASKCLUSTER_SECRET")
     DEFAULTS = {
         "merge": {
+            "autoclassification": {
+                "enabled": False,
+                "test-suite-names": [],
+            },
```
ahal, line 115: Could you also update the docs please?
https://github.com/mozilla/mozci/blob/master/docs/configuration.rst

mozci/task.py
ahal: I kind of feel the same way about the `Push` class, but that's out of scope for now. Maybe we can make a new `classify.py` or `regressions.py` module to start housing this stuff. But leaving it in `task.py` is fine for now.

ahal (changes requested):

mozci/task.py
La0: Done

mozci/configuration.py
La0: Done !

mozci/task.py
```diff
@@ -123,6 +124,19 @@ def is_bad_group(task_id: str, group: str) -> bool:
     return bad_group
 
 
+def is_autoclassifiable(task: Task) -> bool:
+    """Check a task is enabled for auto-classification
+    by applying glob patterns from configuration
+    """
+    if not task.label or not config["autoclassification"]["enabled"]:
```
marco-c, line 131: Maybe we should assert that `task.label` is present

marco-c (approved):

mozci/task.py
La0: Done

marco-c (approved):
Thanks!

PR #705, opened 2022-03-22
Include failure types filtering logic in is_autoclassifiable function

### LINKED ISSUE #685: Add a way to filter based on the type of failure
Follow-up to #679.

Depends on #684.

EvaBardou:
Closes #685

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -124,7 +124,7 @@ def is_bad_group(task_id: str, group: str) -> bool:
     return bad_group
 
 
-def is_autoclassifiable(task: Task) -> bool:
+def is_autoclassifiable(task: TestTask) -> bool:
     """Check a task is enabled for auto-classification
     by applying glob patterns from configuration
     """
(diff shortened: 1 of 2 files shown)


mozci/task.py
```diff
@@ -133,9 +133,25 @@ def is_autoclassifiable(task: Task) -> bool:
...
+        for test_and_type in group
+    ]
+
+    return (
+        any(
+            fnmatch.fnmatch(task.label, pattern)
+            for pattern in config["autoclassification"]["test-suite-names"]
+        )
+        and len(flat_failure_types) == 1
+        and flat_failure_types[0][1].value in filtered_failure_types
```
marco-c, line 154: Nit: what if you replace this with something like:
```Python
and len(task.failure_types) == 1
and task.failure_types[0][1].value in filtered_failure_types
```
so you don't have to create `flat_failure_types`.

marco-c (approved):

mozci/task.py
EvaBardou: `task.failure_types` are grouped by `group` first, that's why I used `flat_failure_types` to get rid of this level

marco-c:
```Python
and len(task.failure_types) == 1 and len(task.failure_types[0]) == 1
and task.failure_types[0][0][1].value in filtered_failure_types
```
should work then, but maybe not so readable...

PR #717, opened 2022-04-04
Display Push author and partial date in classify notifications

### LINKED ISSUE #716: Add push author and push time to the notifications
This data can be useful to sheriffs to ping the author of the push and to identify where the push is in their Treeherder view.

EvaBardou:
Closes #716

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -307,9 +307,9 @@ class ClassifyCommand(Command):
             previous == PushStatus.BAD
             and current in (PushStatus.GOOD, PushStatus.UNKNOWN)
         ):
-            formatted_date = datetime.datetime.fromtimestamp(push.date).strftime(
-                "Pushed at %H hours %M minutes"
-            )
+            formatted_date = datetime.datetime.fromtimestamp(
+                push.date, tz=datetime.timezone.utc
+            ).strftime("Pushed at %H hours %M minutes %z")
             email_content = EMAIL_PUSH_EVOLUTION.format(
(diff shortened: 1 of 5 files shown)


mozci/console/commands/push.py
```diff
@@ -49,6 +49,8 @@
 # Push {push.id} evolved from {previous} to {current}
 
 Rev: [{push.rev}](https://treeherder.mozilla.org/jobs?repo={branch}&revision={push.rev})
+Author: {push.author}
+Date: {date}
```
marco-c, line 53:
```suggestion
Time: {date}
```

mozci/console/commands/push.py
```diff
@@ -305,10 +307,14 @@ def _list_tasks(tasks):
             previous == PushStatus.BAD
             and current in (PushStatus.GOOD, PushStatus.UNKNOWN)
         ):
+            formatted_date = datetime.datetime.fromtimestamp(
+                push.date, tz=datetime.timezone.utc
+            ).strftime("Pushed at %H hours %M minutes %z")
```
marco-c, line 312:
```suggestion
            ).strftime("%H:%M")
```

@Archaeopteryx is treeherder always using UTC timezone?
If so, we should always use UTC timezone here too

Archaeopteryx: Makes sense if we cannot display it in the timezone used by the client.

Archaeopteryx: Please also add the seconds to the timestamp, there can be more than one push per minute.

PR #736, opened 2022-04-27
Add a new command to check backfills completion on last pushes

### LINKED ISSUE #728: Send a notification in the Matrix channel when a backfill is finished
Sheriffs often trigger backfills and have to periodically check if the tasks triggered by the backfill are finished. It would be nice to implement a notification to the Matrix channel to notify them instead, so they don't have to periodically check.

EvaBardou:
Closes #728

### DIFF
--- a/mozci/console/commands/check_backfills.py
+++ b/mozci/console/commands/check_backfills.py
@@ -29,7 +29,7 @@ class CheckBackfillsCommand(Command):
 
     check-backfills
         {--branch=autoland : Branch the pushes belongs to (e.g autoland, try, etc).}
-        {--nb-pushes=100 : Number of recent pushes to retrieve for the check.}
+        {--nb-pushes=40 : Number of recent pushes to retrieve for the check.}
         {--environment=testing : Environment in which the analysis is running (testing, production, ...)}
 
     The command will execute the following workflow:
--- a/mozci/console/commands/check_backfills.py
+++ b/mozci/console/commands/check_backfills.py
@@ -1,0 +1,173 @@
+# -*- coding: utf-8 -*-
+import os
+from collections import namedtuple
+from itertools import groupby
+
+import requests
+from cleo import Command
+from loguru import logger
+
+from mozci import config
+from mozci.push import make_push_objects
+from mozci.util.taskcluster import (
+    COMMUNITY_TASKCLUSTER_ROOT_URL,
+    index_current_task,
+    list_dependent_tasks,
+    list_indexed_tasks,
+    notify_matrix,
+)
+
+BackfillTask = namedtuple("BackfillTask", ["task_id", "group_id", "state"])
+
+NOTIFICATION_BACKFILL_GROUP_COMPLETED = (
+    "Backfill group {group_id} for push {push.branch}/{push.rev} completed."
+)
+
+
+class CheckBackfillsCommand(Command):
+    """
+    Check if backfills on last pushes are finished and notify Sheriffs when they are.
+
+    check-backfills
+        {--branch=autoland : Branch the pushes belongs to (e.g autoland, try, etc).}
+        {--nb-pushes=100 : Number of recent pushes to retrieve for the check.}
+        {--environment=testing : Environment in which the analysis is running (testing, production, ...)}
+
+    The command will execute the following workflow:
+      1. Retrieve the last <--nb-pushes> Pushes on branch <--branch>
+      => Then for each Push:
+      2. Check if it has any associated actions triggered by Treeherder, using Taskcluster indexation
+      3. Find potential backfill tasks
+      4. Group them by backfill groups
+      => For each backfill group on the Push:
+      5. Check if all backfill tasks in this group are completed
+      6. If so (and if the notification wasn't already sent), send a notification on Matrix alerting that this backfill group is completed
+      7. Add the current task in a dedicated index to avoid sending multiple times the same notification
+    """
+
+    def handle(self) -> None:
+        branch = self.option("branch")
+        environment = self.option("environment")
+        matrix_room = config.get("matrix-room-id")
+        current_task_id = os.environ.get("TASK_ID")
+
+        try:
+            nb_pushes = int(self.option("nb-pushes"))
+        except ValueError:
+            self.line("<error>Provided --nb-pushes should be an int.</error>")
+            exit(1)
+
+        self.line("<comment>Loading pushes...</comment>")
+        self.pushes = make_push_objects(nb=nb_pushes, branch=branch)
+        nb_pushes = len(self.pushes)
+
+        for index, push in enumerate(self.pushes, start=1):
+            self.line(
+                f"<comment>Processing push {index}/{nb_pushes}: {push.push_uuid}</comment>"
+            )
+            backfill_tasks = []
+
+            try:
+                indexed_tasks = list_indexed_tasks(
+                    f"gecko.v2.{push.branch}.revision.{push.rev}.taskgraph.actions"
+                )
+            except requests.exceptions.HTTPError as e:
+                self.line(
+                    f"<error>Couldn't fetch indexed tasks on push {push.push_uuid}: {e}</error>"
+                )
+                continue
+
+            for indexed_task in indexed_tasks:
+                task_id = indexed_task["taskId"]
+                try:
+                    children_tasks = list_dependent_tasks(task_id)
+                except requests.exceptions.HTTPError as e:
+                    self.line(
+                        f"<error>Couldn't fetch dependent tasks of indexed task {task_id} on push {push.push_uuid}: {e}</error>"
+                    )
+                    continue
+
+                for child_task in children_tasks:
+                    task_action = (
+                        child_task.get("task", {}).get("tags", {}).get("action", "")
+                    )
+                    # We are looking for the Treeherder symbol because Sheriffs are
+                    # only interested in backfill-tasks holding the '-bk' suffix in TH
+                    th_symbol = (
+                        child_task.get("task", {})
+                        .get("extra", {})
+                        .get("treeherder", {})
+                        .get("symbol", "")
+                    )
+                    status = child_task.get("status", {})
+                    if task_action == "backfill-task" and th_symbol.endswith("-bk"):
+                        assert status.get(
+                            "taskId"
+                        ), "Missing taskId attribute in backfill task status"
+                        assert status.get(
+                            "taskGroupId"
+                        ), "Missing taskGroupId attribute in backfill task status"
+                        assert status.get(
+                            "state"
+                        ), "Missing state attribute in backfill task status"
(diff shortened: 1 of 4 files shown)


mozci/util/taskcluster.py
```diff
@@ -81,6 +83,55 @@ def find_task_id(index_path, use_proxy=False, root_url=PRODUCTION_TASKCLUSTER_RO
...
+    namespace, use_proxy=False, root_url=PRODUCTION_TASKCLUSTER_ROOT_URL
+):
+    url = get_indexed_tasks_url(namespace, root_url=root_url)
+    tasks = []
+    token = False
+    # Support pagination using continuation token
+    while token is not None:
+        extra_params = "?" + urlencode({"continuationToken": token}) if token else ""
+        results = _do_request(url + extra_params).json()
+        tasks += results.get("tasks")
```
La0, line 129: You could use `yield` here instead

mozci/util/taskcluster.py
```diff
@@ -91,6 +142,26 @@ def get_task(task_id, use_proxy=False):
     return queue.task(task_id)
 
 
+def get_dependent_tasks_url(task_id):
```
La0, line 145: Would be more consistent to support passing the TC root url as in `get_indexed_tasks_url`

mozci/util/taskcluster.py
```diff
@@ -81,6 +83,55 @@ def find_task_id(index_path, use_proxy=False, root_url=PRODUCTION_TASKCLUSTER_RO
     return response.json()["taskId"]
 
 
+def insert_task(
```
La0, line 86: You could directly create `index_current_task` here:
- do not pass `task_id` and use directly `os.environ["TASK_ID"]` or raising an error when it's not available
- `use_proxy` seems unused ?

mozci/console/commands/push.py
```diff
@@ -946,6 +955,107 @@ def list_classification_tasks(self, group_id):
...
+    check-backfills
+        {--branch=autoland : Branch the pushes belongs to (e.g autoland, try, etc).}
+        {--nb-pushes=100 : Number of recent pushes to retrieve for the check.}
+        {--environment=testing : Environment in which the analysis is running (testing, production, ...)}
+    """
+
+    def handle(self) -> None:
+        branch = self.option("branch")
+        environment = self.option("environment")
+        matrix_room = config.get("matrix-room-id", None)
```
La0, line 971: Nit: No need to specify `None` it's the default case

mozci/console/commands/push.py
```diff
@@ -946,6 +955,107 @@ def list_classification_tasks(self, group_id):
...
+                        continue
+
+                    # Populating the index with the current task to prevent sending the following notification once again
+                    insert_task(
+                        index_path,
+                        os.environ.get("TASK_ID"),
+                        root_url=COMMUNITY_TASKCLUSTER_ROOT_URL,
+                    )
+
+                    if not matrix_room:
```
La0, line 1046: Add a logger.info here saying that a notification should be sent for group XXX.

La0: You could even `print` the notification

mozci/console/commands/push.py
```diff
@@ -946,6 +955,107 @@ def list_classification_tasks(self, group_id):
...
+                if all(task["state"] == "completed" for task in tasks):
+                    index_path = f"project.mozci.check-backfill.{environment}.{push.branch}.{push.rev}.{group_id}"
+                    already_indexed = list_indexed_tasks(
+                        index_path, root_url=COMMUNITY_TASKCLUSTER_ROOT_URL
+                    )
+                    if already_indexed:
+                        continue
+
+                    # Populating the index with the current task to prevent sending the following notification once again
+                    insert_task(
```
La0, line 1040: Only run that when TASK_ID is available (it will only be available on Taskcluster, not while running locally)

mozci/console/commands/push.py
```diff
@@ -946,6 +955,107 @@ def list_classification_tasks(self, group_id):
             config.cache.add(cache_key, tasks, int(config["cache"]["retention"]))
 
 
+class CheckBackfillsCommand(Command):
+    """
+    Check if backfills on last pushes are finished and notify Sheriffs when they are.
```
La0, line 960: This docstring should be extended to list all the stages in the workflow:
- list NB pushes on branch
- check if they have any actions triggered by Treeherder, using TC index
- find potential backfill tasks
- check if all tasks in the group of a backfill tasks are completed
- send a notification
- index current task to avoid retrials

mozci/console/commands/push.py
La0, line 1040: The indexation must happen AFTER the notification: we only want to track notifications to avoid re-sending them

mozci/console/commands/push.py
```diff
@@ -946,6 +955,107 @@ def list_classification_tasks(self, group_id):
...
+
+            def group_key(task):
+                return task["taskGroupId"]
+
+            backfill_tasks = sorted(backfill_tasks, key=group_key)
+            for group_id, value in groupby(backfill_tasks, group_key):
+                tasks = list(value)
+                if all(task["state"] == "completed" for task in tasks):
+                    index_path = f"project.mozci.check-backfill.{environment}.{push.branch}.{push.rev}.{group_id}"
+                    already_indexed = list_indexed_tasks(
```
La0, line 1033: You need to catch exceptions there, when no index is found

mozci/console/commands/push.py
```diff
@@ -946,6 +955,107 @@ def list_classification_tasks(self, group_id):
...
+                        child_task.get("task", {})
+                        .get("extra", {})
+                        .get("treeherder", {})
+                        .get("symbol", "")
+                    )
+                    if task_action == "backfill-task" and th_symbol.endswith("-bk"):
+                        status = child_task.get("status", {})
+                        backfill_tasks.append(
+                            {
+                                "taskId": status.get("taskId"),
```
La0, line 1019: I would prefer a namedtuple here, and stricter checks: the workflow cannot work if any of these information is missing.

So raise an exception when a part is missing

mozci/console/commands/push.py
```diff
@@ -946,6 +955,107 @@ def list_classification_tasks(self, group_id):
...
+        except ValueError:
+            self.line("<error>Provided --nb-pushes should be an int.</error>")
+            exit(1)
+
+        self.line("<comment>Loading pushes...</comment>")
+        self.pushes = make_push_objects(nb=nb_pushes, branch=branch)
+
+        for push in self.pushes:
+            backfill_tasks = []
+
```
La0, line 984: Add a logger.debug showing the progression over all pushes:

```
Processing push 1/100: <push.id>
```

mozci/console/commands/push.py
La0: My bad, it's already caught.

La0:
As discussed with @marco-c, please update the command to expose it at root level (as `mozci check-backfills`)

mozci/util/taskcluster.py
EvaBardou: Some calls have it in this file, some don't. But I will add the `root_url` anyway, it's more versatile.

mozci/util/taskcluster.py
EvaBardou: I reused some existing code from: https://github.com/mozilla/mozci/blob/master/mozci/console/commands/push.py#L915

Should I make this change anyway?

mozci/util/taskcluster.py
EvaBardou: Sure, I implemented it like this so that the function is easily reusable, but I can refactor it.

mozci/util/taskcluster.py
La0: Yes because it's a standalone method, it will be cleaner to use :

```python
yield from results.get("tasks", [])
```

mozci/console/commands/check_backfills.py
```diff
@@ -0,0 +1,173 @@
...
+                                status["taskId"], status["taskGroupId"], status["state"]
+                            )
+                        )
+                    else:
+                        logger.debug(
+                            f"Skipping non-backfill task {status.get('taskId')}"
+                        )
+
+            def group_key(task):
+                return task.group_id
```
marco-c, line 124: Backfill tasks belonging to the same backfill action might be running on different task groups, so if we group by task group we will notify at wrong times.

We should instead group backfills by the treeherder symbol, as all backfill tasks related to a given backfill have the same symbol.

Right @Archaeopteryx?

mozci/console/commands/check_backfills.py
```diff
@@ -0,0 +1,173 @@
...
+            # Sorting backfill tasks by their Taskcluster backfill group_id
+            backfill_tasks = sorted(backfill_tasks, key=group_key)
+            # Grouping ordered backfill tasks by their associated Taskcluster backfill group
+            for group_id, value in groupby(backfill_tasks, group_key):
+                tasks = list(value)
+                if all(task.state == "completed" for task in tasks):
+                    index_path = f"project.mozci.check-backfill.{environment}.{push.branch}.{push.rev}.{group_id}"
+                    already_indexed = [
+                        indexed
+                        for indexed in list_indexed_tasks(
```
marco-c, line 135: Nit: In this case we should only ever have one indexed task, right? Maybe we can simply check for the existence of the index instead of listing?

mozci/console/commands/check_backfills.py
Archaeopteryx: Yes, the notification should only be sent if the tasks in all groups have finished.

EvaBardou: FYI: With this refactoring, the current task will be indexed under `project.mozci.check-backfill.<environment>.<branch>.<revision>.<treeherder_symbol>` when a notification is sent

mozci/console/commands/check_backfills.py
```diff
@@ -0,0 +1,167 @@
...
+    find_task_id,
+    index_current_task,
+    list_dependent_tasks,
+    list_indexed_tasks,
+    notify_matrix,
+)
+
+BackfillTask = namedtuple("BackfillTask", ["task_id", "th_symbol", "state"])
+
+NOTIFICATION_BACKFILL_GROUP_COMPLETED = "Backfill tasks associated to the Treeherder symbol {th_symbol} for push {push.branch}/{push.rev} are all completed."
```
marco-c, line 23: @Archaeopteryx does this format for the notification look good to you? Can you suggest improvements if you think some are needed?

Archaeopteryx: Looks good to me.

marco-c (approved):

marco-c:
@La0 we can merge this if you're OK with the latest iteration of the patch.

La0 (approved):
Code & output are way better :100:

PR #738, opened 2022-04-28
Switch to using index.buildUrl and manual requests.get instead of relying on the taskcluster library to get the artifact

### LINKED ISSUE #737: Multiple notifications sent for the same classification and the same push
Today we've seen dozens of `Push 161849 evolved from no classification to BAD` notifications, with no difference between each of them.
For some reason mozci is failing to check if a notification was already sent for the push.

marco-c:
The taskcluster library is failing on 303.

Fixes #737

### DIFF
--- a/mozci/data/sources/taskcluster/__init__.py
+++ b/mozci/data/sources/taskcluster/__init__.py
@@ -150,7 +150,7 @@ class TaskclusterSource(DataSource):
             )
 
         try:
-            r = requests.get(url)
+            r = requests.get(url, allow_redirects=True)
             r.raise_for_status()
         except requests.exceptions.HTTPError as e:
             raise ContractNotFilled(
--- a/mozci/data/sources/taskcluster/__init__.py
+++ b/mozci/data/sources/taskcluster/__init__.py
@@ -132,6 +132,8 @@ class TaskclusterSource(DataSource):
             else f"project.mozci.{environment}.classification"
         )
 
+        # We use buildUrl and manual requests.get instead of directly findArtifactFromTask from the taskcluster library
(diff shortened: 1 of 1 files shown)


mozci/data/sources/taskcluster/__init__.py
```diff
@@ -147,7 +148,17 @@ def run_push_existing_classification(self, branch, rev, environment):
             )
 
         try:
-            return response["push"]["classification"]
+            r = requests.get(url)
```
La0, line 151: I would either add `allow_redirects=Trues` or add a comment explaining we don't use the internal method because it may fail on redirections

La0 (approved):
Now Works for me:

```
 python crash.py
2022-04-28 12:41:04.400 | DEBUG    | mozci.configuration:__init__:87 - Active cache store file with conf {'driver': 'file', 'path': '/home/bastien/dev/mozilla/mozci.cache'}
12:41:04.427 | INFO     | mozci.data.base:50  - Sources selected, in order of priority: ('hgmo', 'taskcluster', 'treeherder_client', 'errorsummary', 'bugbug').
PushStatus.BAD
```

for 

```python
from mozci.push import Push

push = Push("a901a6c263672b4a625a7d5948c8a70d330ca7e6")
print(push.get_existing_classification("production"))
```

which was failing before on master

mozci/data/sources/taskcluster/__init__.py
marco-c: Ah, that's the name, I tried using `follow_redirects` but the parameter didn't exist so I simply lazily abandoned the idea of adding it :P

marco-c: Done both with ee83b1a and 765b355.

PR #744, opened 2022-05-06
Allow to set 'classify' parameters through config in Classify and ClassifyEval commands

### LINKED ISSUE #735: Define autoclassification parameters in the config secret
The parameters to https://github.com/mozilla/mozci/blob/b466f55213cfe8b99b10f84610e2e4272c09e559/mozci/push.py#L1218-L1227 should be defined in the secret so we can periodically check the best ones and simply update the secret instead of having to modify the code and push a new release.

EvaBardou:
Closes #735

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -6,7 +6,7 @@ import os
 import re
 import traceback
 from inspect import signature
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 from urllib.parse import urlencode
 
 import arrow
@@ -59,6 +59,8 @@ Time: {date}
 - {real_failures}
 
 """
+
+TWO_INTS_TUPLE_REGEXP = r"^\((\d+), ?(\d+)\)$"
 
 
 class PushTasksCommand(Command):
@@ -130,10 +132,19 @@ def check_type(parameter_type, hint, value):
     try:
         if parameter_type == bool:
             parameter = value not in [False, 0, "0", "False", "false", "f"]
+        elif parameter_type == Optional[Tuple[int, int]]:
+            match = re.match(TWO_INTS_TUPLE_REGEXP, value)
+
+            if not match or len(match.groups()) != 2:
+                raise ValueError
+
+            parameter = tuple([int(number) for number in match.groups()])
         else:
             parameter = parameter_type(value)
     except ValueError:
-        raise Exception(f"Provided {hint} should be a {parameter_type.__name__}.")
+        raise Exception(
+            f"Provided {hint} should be a {parameter_type.__name__ if hasattr(parameter_type, '__name__') else parameter_type}."
+        )
 
     return parameter
 
@@ -411,13 +422,20 @@ class ClassifyEvalCommand(Command):
             self.option("rev"),
         )
 
+        option_names = [
+            "intermittent-confidence-threshold",
+            "real-confidence-threshold",
+            "use-possible-regressions",
+            "unknown-from-regressions",
+            "consider-children-pushes-configs",
+            "cross-config-counts",
+            "consistent-failures-counts",
+        ]
         if self.option("recalculate"):
             classify_parameters = retrieve_classify_parameters(self.option)
-        elif self.option("intermittent-confidence-threshold") or self.option(
-            "real-confidence-threshold"
-        ):
+        elif any(self.option(name) for name in option_names):
             self.line(
-                "<error>--recalculate isn't set, you shouldn't provide either --intermittent-confidence-threshold nor --real-confidence-threshold attributes.</error>"
+                f"<error>--recalculate isn't set, you shouldn't provide --{', --'.join(option_names)} CLI options.</error>"
             )
             return
 
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -33,6 +33,16 @@ from mozci.util.taskcluster import (
     notify_matrix,
 )
 
+CLASSIFY_PARAMETERS = [
+    "intermittent_confidence_threshold",
+    "real_confidence_threshold",
+    "use_possible_regressions",
+    "unknown_from_regressions",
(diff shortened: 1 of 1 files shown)


mozci/console/commands/push.py
```diff
@@ -124,6 +134,40 @@ def classify_commands_pushes(
...
+    if high_conf_option is not None:
+        if classify_parameters.get(parameter_name) is not None:
+            raise Exception(
+                f"You should provide either --high-confidence CLI option OR '{parameter_name}' in the config secret not both."
+            )
+
+        try:
+            classify_parameters[parameter_name] = float(high_conf_option)
+        except ValueError:
+            raise Exception("Provided --high-confidence should be a float.")
```
marco-c, line 166: Could we have a list of options through `Command` so we don't have to check them one by one manually?

marco-c: If it isn't feasible, we could define options for all of the values in CLASSIFY_PARAMETERS, and then use CLASSIFY_PARAMETERS as a list (checking that no items in CLASSIFY_PARAMETERS are both in config and passed as parameters)

marco-c: I was asking to avoid hard-coding the option names here, but check if it is possible to get all option names from the `Command` instance.
So you'd do something like:
```
    for parameter_name in self.option_names():
        if config.get(parameter_name) is not None and self.options(parameter_name) is not None:
            raise Exception(
                f"You should provide either --{parameter_name} CLI option OR '{parameter_name}' in the config secret not both."
            )
```

If that isn't feasible, then do something like:
```
    for parameter_name in CLASSIFY_PARAMETERS:
        if config.get(parameter_name) is not None and self.options(parameter_name) is not None:
            raise Exception(
                f"You should provide either --{parameter_name} CLI option OR '{parameter_name}' in the config secret not both."
            )
```
possibly defining all CLASSIFY_PARAMETERS as options (this last point is not required as we are not using them right now like you said, but nice to have as we might want to use them in the future for local runs of classify eval, if we want to run multiple in parallel and so using config isn't feasible).

marco-c: They'd just be ignored as they are not also configuration options

EvaBardou: I pushed a new commit where I rely on the `classify` function signature instead of a constant to get the supported parameters. I also added support for all parameters as CLI options.

For each parameter, one can :
- define nothing, then the default value specified in the `classify` signature will be used.
- set the parameter via config **AND** via a CLI option, then an error will be raised.
- define the parameter via the config, it will be checked that the provided value is castable to the actual type of the parameter.
- define the parameter via CLI option, it will be checked that the provided value is castable to the actual type of the parameter.

EvaBardou:
@marco-c I pushed a new commit where I rely on the `classify` function signature instead of a constant to get the supported parameters. I also added support for all parameters as CLI options.

For each parameter, one can :
- define nothing, then the default value specified in the `classify` signature will be used.
- set the parameter via config **AND** via a CLI option, then an error will be raised.
- define the parameter via the config, it will be checked that the provided value is castable to the actual type of the parameter.
- define the parameter via CLI option, it will be checked that the provided value is castable to the actual type of the parameter.

mozci/console/commands/push.py
```diff
@@ -345,8 +386,13 @@ class ClassifyEvalCommand(Command):
...
         {--to-date= : Upper bound of the push range (as a date in yyyy-mm-dd format or a human expression like "1 days ago"), defaults to now.}
-        {--medium-confidence= : If recalculate parameter is set, medium confidence threshold used to classify the regressions.}
-        {--high-confidence= : If recalculate parameter is set, high confidence threshold used to classify the regressions.}
+        {--intermittent-confidence-threshold= : Medium confidence threshold used to classify the regressions.}
+        {--real-confidence-threshold= : High confidence threshold used to classify the regressions.}
+        {--use-possible-regressions= : Use possible regressions while classifying the regressions.}
+        {--unknown-from-regressions= : Unknown from regressions while classifying the regressions.}
+        {--consider-children-pushes-configs= : Consider children pushes configs while classifying the regressions.}
+        {--cross-config-counts= : Cross-config counts used to classify the regressions.}
+        {--consistent-failures-counts= : Consistent failures counts used to classify the regressions.}
```
marco-c, line 395: The `*-counts` options don't seem to work from the CLI because they are tuples. We can handle it in a follow-up if you want, it's not a big priority.

mozci/console/commands/push.py
```diff
@@ -366,31 +412,12 @@ def handle(self) -> None:
...
-                    else 0.9
-                )
-            except ValueError:
-                self.line(
-                    "<error>Provided --high-confidence should be a float.</error>"
-                )
-                exit(1)
-        elif self.option("medium-confidence") or self.option("high-confidence"):
+            classify_parameters = retrieve_classify_parameters(self.option)
+        elif self.option("intermittent-confidence-threshold") or self.option(
```
marco-c, line 416: Now that we have more options, we should fail if recalculate isn't set and any of the classify options are used

marco-c (approved):
LGTM! There are two things to fix, but they don't block landing this.
If you want to fix them later, feel free to merge this and file two follow-up issues.

mozci/console/commands/push.py
```diff
@@ -365,32 +422,20 @@ def handle(self) -> None:
...
 
+        option_names = [
+            "intermittent-confidence-threshold",
+            "real-confidence-threshold",
+            "use-possible-regressions",
+            "unknown-from-regressions",
+            "consider-children-pushes-configs",
+            "cross-config-counts",
+            "consistent-failures-counts",
+        ]
```
marco-c, line 433: Nit: we could use inspect.signature here too to avoid listing them all

marco-c (approved):

mozci/console/commands/push.py
EvaBardou: Change pushed :smile:

PR #748, opened 2022-05-06
Use taskcluster library to index check-backfill tasks

marco-c:
So we can use the Taskcluster proxy easily and so we actually use the scopes we are assigned


marco-c:
I tested this manually and it is working, so I'll just merge it.

PR #758, opened 2022-05-12
Add a filter on the task name in the check-backfills notification

### LINKED ISSUE #754: In the backfill finished notification, add a filter on task name
As @Archaeopteryx mentioned on Matrix, the visible tasks should be filtered based on the task name: 1. drop the chunk number at the end. 2. drop "-e10s".
For example for `Backfill tasks associated to the Treeherder symbol 4-a3024e8b2a5-bk for push [autoland/4a62147cede6c8a2f9a516770bbea5c8ce7c5118](https://treeherder.mozilla.org/jobs?repo=autoland&tochange=4a62147cede6c8a2f9a516770bbea5c8ce7c5118&fromchange=292df8ed886d5b62860f0d88054c5cbb7e71aea1) are all in a final state.`: https://treeherder.mozilla.org/jobs?repo=autoland&tochange=4a62147cede6c8a2f9a516770bbea5c8ce7c5118&fromchange=292df8ed886d5b62860f0d88054c5cbb7e71aea1&searchStr=test-android-em-7.0-x86_64-lite-qr%2Fopt-geckoview-mochitest-plain-nofis.

EvaBardou:
Closes #754

### DIFF
--- a/mozci/console/commands/check_backfills.py
+++ b/mozci/console/commands/check_backfills.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import os
+import re
 from collections import namedtuple
 from itertools import groupby
 from typing import Any, Dict
@@ -175,11 +176,14 @@ class CheckBackfillsCommand(Command):
                 )
                 parents = None
 
+            cleaned_label = re.sub(
+                r"(-e10s|-1proc)?(-\d+)?$", "", all_backfill_tasks.pop().label
+            )
             notification = NOTIFICATION_BACKFILL_GROUP_COMPLETED.format(
                 th_symbol=th_symbol,
                 push=newest_push,
                 fromchange=f"&fromchange={parents[-1].rev}" if parents else "",
-                searchstr=f"&searchStr={all_backfill_tasks.pop().label.split('-e10s')[0]}",
+                searchstr=f"&searchStr={cleaned_label}",
             )
 
             if not matrix_room:


mozci/console/commands/check_backfills.py
```diff
@@ -172,9 +178,8 @@ def group_key(task):
             notification = NOTIFICATION_BACKFILL_GROUP_COMPLETED.format(
                 th_symbol=th_symbol,
                 push=newest_push,
-                extra_with_twentieth_parent=f"&fromchange={parents[-1].rev}"
-                if parents
-                else "",
+                fromchange=f"&fromchange={parents[-1].rev}" if parents else "",
+                searchstr=f"&searchStr={all_backfill_tasks.pop().label.split('-e10s')[0]}",
```
marco-c, line 182: @Archaeopteryx @jmaher @ahal do you think this works for all task labels? (the goal is removing `e10s` and the chunk number from the label to filter tasks on treeherder as @Archaeopteryx suggested)

Archaeopteryx: This e.g. won't cover builds. Treeherder's [code for this](https://github.com/mozilla/treeherder/blob/bf87a56cfbe1b3ec0aad0a0fb8bf8229eddb31ea/ui/helpers/job.js#L233) also doesn't work for every task.
```
all_backfill_tasks.pop().label..replace(/(-e10s)?(-\d+)?$/, "")
```
would show all chunks in that group for that test suite.

jmaher: I think we are backfilling tests, not builds.  the two exceptions I can think of is jittest and mochitest-chrome where we do not run those in e10s, but 1proc and we chunk them:
https://treeherder.mozilla.org/jobs?repo=mozilla-central&tier=1%2C2%2C3&searchStr=jittest&revision=b1a6d489400499979f16cc982901f78ac504bc77

https://treeherder.mozilla.org/jobs?repo=mozilla-central&selectedTaskRun=FeW_ngbhS0uNBGyZmCC1lg.0&tier=1%2C2%2C3&searchStr=1proc%2Cchrome&revision=b1a6d489400499979f16cc982901f78ac504bc77

marco-c: All right, so we should exclude both e10s and 1proc, so I guess `all_backfill_tasks.pop().label.replace(/(-e10s)?(-1proc)?(-\d+)?$/, "")` should work.

PR #761, opened 2022-05-18
Warn about pushes that are backed-out and where all failures on the push itself and its children are marked as intermittent

### LINKED ISSUE #739: Add some additional monitoring to ensure autoclassification is behaving correctly
E.g. if a push is backed out and we have all failures on the push and its children classified as intermittent, then there's something wrong and we should manually check the autoclassifications.

- [x] Warn about pushes that are backed-out and where all failures on the push itself and its children are marked as intermittent (#761)
- [x] Warn if there is a classification that references a revision that does not exist (#779)
- [x] Warn if there is a group that is failing in a set of pushes and marked sometimes as intermittent and sometimes as fixed by commit or sometimes as fixed by a backout and sometimes fixed by another backout (#781)
- [x]  Build a graph that shows the number of failures classified as intermittent (divided by number of pushes and by number of tasks) over time: [Intermittent failures/100 adjacent pushes](https://sql.telemetry.mozilla.org/queries/86531#214296)

- [x]  Build a graph of average number of intermittents per push or per 100 jobs: see above

- [x]  Show number of retriggers and number of backfills we perform (divided by number of pushes and by number of tasks) over time: [overall count of additional tasks](https://sql.telemetry.mozilla.org/queries/86793#215014) - includes those requested by performance sheriffs
- [x] Warn if there is a failure that starts happening on a push that comes before the push that is backed-out and which is marked as "fixed by commit" of the backout of the second push (https://github.com/mozilla/mozci/blob/6cf1980876d9c48562e983728c8f253ca9cb7073/mozci/push.py#L564) (#783)

EvaBardou:
Linked to #739

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -567,6 +567,21 @@ class ClassifyEvalCommand(Command):
                 if ever_classified_as_cause:
                     break
 
+            if not ever_classified_as_cause:
+                for (
+                    _,
+                    _,
+                    candidate_regressions,
+                    classified_as_cause,
+                ) in push._iterate_failures("group", 50):
+                    ever_classified_as_cause = any(
+                        result is True
+                        for name in candidate_regressions.keys()
+                        for result in classified_as_cause[name]
+                    )
+                    if ever_classified_as_cause:
+                        break
+
             if push.backedout and not ever_classified_as_cause:
                 self.line(
                     f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -552,20 +552,24 @@ class ClassifyEvalCommand(Command):
                     )
                     self.errors[push] = e
 
-            all_pushes = set(
-                [push] + [child for child in push._iterate_children(max_depth=50)]
-            )
-            all_failures_classifications = [
-                set([c for c, _ in group.classifications])
-                for p in all_pushes
-                for group in p.group_summaries.values()
-            ]
-            if push.backedout and all(
-                failure_classifications == {"intermittent"}
-                for failure_classifications in all_failures_classifications
-            ):
+            ever_classified_as_cause = False
+            for (
+                _,
+                _,
+                candidate_regressions,
+                classified_as_cause,
+            ) in push._iterate_failures("label", 50):
+                ever_classified_as_cause = any(
+                    result is True
+                    for name in candidate_regressions.keys()
+                    for result in classified_as_cause[name]
+                )
+                if ever_classified_as_cause:
+                    break
+
+            if push.backedout and ever_classified_as_cause:
                 self.line(
-                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent by Sheriffs</comment>"
+                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent</comment>"
                 )
                 # Build the graph, etc
 
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -552,6 +552,23 @@ class ClassifyEvalCommand(Command):
                     )
                     self.errors[push] = e
 
+            all_pushes = set(
+                [push] + [child for child in push._iterate_children(max_depth=50)]
+            )
+            all_failures_classifications = [
+                set([c for c, _ in group.classifications])
+                for p in all_pushes
+                for group in p.group_summaries.values()
+            ]
+            if push.backedout and all(
+                failure_classifications == {"intermittent"}
+                for failure_classifications in all_failures_classifications
+            ):
+                self.line(
+                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent by Sheriffs</comment>"
+                )
+                # Build the graph, etc
+
             # Advance the overall progress bar
             progress.advance()
 
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -567,9 +567,9 @@ class ClassifyEvalCommand(Command):
                 if ever_classified_as_cause:
                     break
 
-            if push.backedout and ever_classified_as_cause:
+            if push.backedout and not ever_classified_as_cause:
                 self.line(
-                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent</comment>"
+                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
                 )
                 # Build the graph, etc
 
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -586,7 +586,6 @@ class ClassifyEvalCommand(Command):
                 self.line(
                     f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
                 )
-                # Build the graph, etc
 
             # Advance the overall progress bar
             progress.advance()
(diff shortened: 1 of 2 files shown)


mozci/console/commands/push.py
```diff
@@ -551,6 +551,23 @@ def handle(self) -> None:
...
+                set([c for c, _ in group.classifications])
+                for p in all_pushes
+                for group in p.group_summaries.values()
+            ]
+            if push.backedout and all(
+                failure_classifications == {"intermittent"}
+                for failure_classifications in all_failures_classifications
+            ):
+                self.line(
+                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent by Sheriffs</comment>"
```
marco-c, line 567:
```suggestion
                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent</comment>"
```

We want to warn when classifications from mozci are all intermittent

mozci/console/commands/push.py
```diff
@@ -551,6 +551,23 @@ def handle(self) -> None:
...
+            all_pushes = set(
+                [push] + [child for child in push._iterate_children(max_depth=50)]
+            )
+            all_failures_classifications = [
+                set([c for c, _ in group.classifications])
+                for p in all_pushes
+                for group in p.group_summaries.values()
+            ]
+            if push.backedout and all(
+                failure_classifications == {"intermittent"}
```
marco-c, line 563: We should also exclude failures classified as "fixed by commit", where the cause is not the push itself. You can use the `_is_classified_as_cause` function from the Push class for that.

mozci/console/commands/push.py
EvaBardou: Are you talking about autoclassifications or all `MozciClassification` stored in Treeherder?

EvaBardou: > > where all failures on the push itself and its children are marked as intermittent;
>
> Eva - Should I use the max_depth attribute to list children (so we don't end up fetching 500+ pushes if we are evaluating a really old push)?
> Eva - What does marked as intermittent precisely mean? Classified as intermittent by mozci or marked by a Sheriff?

> Joel - my best guess at answering your question is you should use max_depth, Ideally we would start at a given revision and work from there- I would be 50 revisions would be safe enough for 99% of the situations
> Joel - and I am not sure about marked as intermittent I assume it would mean by sheriffs, but as time goes on it would probably need to include both sheriffs and high confidence mozci

We talked about this with Joel on Matrix while you were on PTO and we didn't quite figure it out apparently

mozci/console/commands/push.py
EvaBardou: I don't really understand how I should use the `_is_classified_as_cause` function:
```
def _is_classified_as_cause(self, first_appareance_push, classifications):
```

Do you know what is the `first_appareance_push` here?

Should I check all children group classifications like so?
```
all_pushes = set(
    [push] + [child for child in push._iterate_children(max_depth=50)]
)
all_failures_classifications = [
    set([c for c, _ in group.classifications])
    for p in all_pushes
    for group in p.group_summaries.values()
    if p._is_classified_as_cause(first_appareance_push???, group.classifications)
]
```

mozci/console/commands/push.py
marco-c: Thinking about it more, we could also just check all classifications from Treeherder, no matter where they are coming from (sheriffs or the mozci hook).

marco-c: We just need to take into account both "intermittent" and "autoclassified intermittent" (but if we do the solution with _iterate_failures I proposed below, we can ignore this aspect anyway)

mozci/console/commands/push.py
marco-c: If `ever_classified_as_cause` is False, it means there is no classification that is not "intermittent", so by using it you are basically doing:
> list pushes that are backed-out and where at least one failure on the push itself and its children is not marked as intermittent;
which is equivalent to 
> list pushes that are backed-out and where all failures on the push itself and its children are marked as intermittent;

mozci/console/commands/push.py
```diff
@@ -551,6 +551,27 @@ def handle(self) -> None:
...
+                    result is True
+                    for name in candidate_regressions.keys()
+                    for result in classified_as_cause[name]
+                )
+                if ever_classified_as_cause:
+                    break
+
+            if push.backedout and ever_classified_as_cause:
+                self.line(
+                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent</comment>"
```
marco-c, line 571:
```suggestion
                    f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
```

marco-c:
Could you try running this on a few pushes so we can check if it is behaving correctly? Hopefully we should have no warnings, unless we missed something (I fear we might have to do some changes, to account for test suites with no groups, but let's see...)

marco-c:
> Sure, would 4 days of pushes be enough? Or maybe you were thinking of fewer pushes?

1 day should be enough

EvaBardou:
> 1 day should be enough

Here are the results for 1 day of pushes (46 in total) using the command `mozci push classify-eval --from-date="1 days ago"`:
```
Loading pushes...
Taskcluster task missing for autoland 5718cf939e0d24b8e7324515f8ff9fc1867aeb7b
Taskcluster task missing for autoland a396a1c0d346b1b59535b280dbd52cf3c5ca6331
Push autoland/15e9b4e9070c569c838f85ca8ceccbf4d52080cc was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push
Push autoland/656906f99173736fb0feb52c989de996561fd725 was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push


Failed to fetch classification for 2 out of 46 pushes. Use the '--recalculate' option if you want to generate them yourself.
0 out of 46 pushes weren't backed-out by a sheriff and were classified as BAD.
1 out of 46 pushes were backed-out by a sheriff and were classified as BAD.
41 out of 46 pushes weren't backed-out by a sheriff and were classified as GOOD.
0 out of 46 pushes were backed-out by a sheriff and were classified as GOOD.
0 out of 46 pushes weren't backed-out by a sheriff and were classified as UNKNOWN.
2 out of 46 pushes were backed-out by a sheriff and were classified as UNKNOWN.
```

The info log was printed two times for `autoland/15e9b4e9070c569c838f85ca8ceccbf4d52080cc` and `autoland/656906f99173736fb0feb52c989de996561fd725`

marco-c:
There's something fishy for 15e9b4e9070c569c838f85ca8ceccbf4d52080cc: the group `layout/xul/test/chrome.ini` failed on 656906f99173736fb0feb52c989de996561fd725, which is the second children of 15e9b4e9070c569c838f85ca8ceccbf4d52080cc, and the task was marked as "fixed by commit 27ea3966f4c9ab75a4e755ff4a2f8ee01322a427". 27ea3966f4c9ab75a4e755ff4a2f8ee01322a427 is the backout of 15e9b4e9070c569c838f85ca8ceccbf4d52080cc.

So `ever_classified_as_cause` should be True for 15e9b4e9070c569c838f85ca8ceccbf4d52080cc.

marco-c:
Ah, the solution is simple: we don't have to only iterate through "label" failures, but through "group" failures too.

EvaBardou:
@marco-c I pushed a fix, new results for the same day as before:
```
Loading pushes...
Taskcluster task missing for autoland 5718cf939e0d24b8e7324515f8ff9fc1867aeb7b
Taskcluster task missing for autoland a396a1c0d346b1b59535b280dbd52cf3c5ca6331


Failed to fetch classification for 2 out of 46 pushes. Use the '--recalculate' option if you want to generate them yourself.
0 out of 46 pushes weren't backed-out by a sheriff and were classified as BAD.
1 out of 46 pushes were backed-out by a sheriff and were classified as BAD.
41 out of 46 pushes weren't backed-out by a sheriff and were classified as GOOD.
0 out of 46 pushes were backed-out by a sheriff and were classified as GOOD.
0 out of 46 pushes weren't backed-out by a sheriff and were classified as UNKNOWN.
2 out of 46 pushes were backed-out by a sheriff and were classified as UNKNOWN.
```

marco-c (approved):
Thanks!

PR #764, opened 2022-05-24
Implement ability to perform backfills on tasks

### LINKED ISSUE #743: Implement ability to perform backfills on a push
See also #508, which implemented retriggers.

To trigger backfills, we need to trigger a backfill action (IIRC there are two potential actions, one for a single backfill on a single push and one for a backfill on multiple pushes).
Here's an example of action triggering (in this case "add-new-jobs" action): https://github.com/mozilla/code-review/blob/47d1b78b80e7f2d59b2d4dd3ef95064f1b2d661d/events/code_review_events/bugbug_utils.py#L240.

EvaBardou:
Closes #743

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -304,10 +304,22 @@ class Task:
             },
         )
 
-        logger.info("Backfilling task '{}'".format(self.tags.get("label", "")))
+        tc_firefox_ci_credentials = config.get("taskcluster_firefox_ci", {})
+        client_id = tc_firefox_ci_credentials.get("client_id")
+        access_token = tc_firefox_ci_credentials.get("access_token")
+        assert (
+            client_id and access_token
+        ), "Missing Taskcluster Firefox CI credentials in mozci config secret"
+
         options = taskcluster.optionsFromEnvironment()
         options["rootUrl"] = PRODUCTION_TASKCLUSTER_ROOT_URL
+        options["credentials"] = {
+            "clientId": client_id,
+            "accessToken": access_token,
+        }
         hooks = taskcluster.Hooks(options)
+
+        logger.info("Backfilling task '{}'".format(self.tags.get("label", "")))
         result = hooks.triggerHook(
             backfill_action["hookGroupId"], backfill_action["hookId"], hook_payload
         )
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -12,6 +12,7 @@ from inspect import signature
 from statistics import median
 from typing import Dict, List, NewType, Optional, Tuple
 
+import jsone
 import requests
 import taskcluster
 from loguru import logger
@@ -25,6 +26,7 @@ from mozci.util.taskcluster import (
     find_task_id,
     get_artifact,
     get_task,
+    get_taskcluster_options,
     list_artifacts,
 )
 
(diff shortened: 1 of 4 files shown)


mozci/task.py
```diff
@@ -285,6 +287,31 @@ def _should_retrigger(self, task):
...
+            backfill_action["hookPayload"],
+            context={
+                "taskId": self.id,
+                "taskGroupId": decision_task.id,
+                "input": {"times": 5 if self.classification == "intermittent" else 1},
+            },
+        )
+
+        logger.info("Backfilling task '{}'".format(self.tags.get("label", "")))
+        hooks = taskcluster.Hooks(get_taskcluster_options())
```
marco-c, line 309: The hook is in firefox-ci rather than community-tc, with this you'd be using the second

marco-c: We have an example (exactly opposite) where we have a program running in firefox-ci that wants to trigger a hook on community-tc.
In that case, we define a `community_taskcluster_config` at https://github.com/mozilla/code-review/blob/2a8437bb3c34205c5096d113080775baa13a7cf9/events/code_review_events/__init__.py#L5, we authenticate at https://github.com/mozilla/code-review/blob/2a8437bb3c34205c5096d113080775baa13a7cf9/events/code_review_events/cli.py#L63-L67 with the client ID and access token we have defined in the secrets, and then we get the "hooks" service at https://github.com/mozilla/code-review/blob/2a8437bb3c34205c5096d113080775baa13a7cf9/events/code_review_events/bugbug_utils.py#L73.

We have to do something similar here.

EvaBardou: Oh okay, I didn't pay attention to that, thank you for the heads up!

marco-c: I think it won't work because from community-tc we wouldn't have the permissions to trigger the hook.
Probably that code snippet was written with the idea that we'd be running from firefox-ci.

marco-c: I think it would work if you also add
```Python
options["credentials"] = {
    "clientId": client_id,
    "accessToken": access_token,
}
```
to the options

marco-c: client_id and access_token will have to be defined as secrets

marco-c:
The client is in the mozci secret with the following format:
```yaml
taskcluster_firefox_ci:
  client_id: CLIENT_ID_NAME
  access_token: ACCESS_TOKEN
  ```

**Comment by EvaBardou** [comment]
> ```yaml
> taskcluster_firefox_ci:
>   client_id: CLIENT_ID_NAME
>   access_token: ACCESS_TOKEN
> ```

@marco-c Isn't it supposed to be something like:
```
taskcluster_firefox_ci = { client_id = "CLIENT_ID_NAME", access_token = "ACCESS_TOKEN" }
```

marco-c:
> > ```yaml
> > taskcluster_firefox_ci:
> >   client_id: CLIENT_ID_NAME
> >   access_token: ACCESS_TOKEN
> > ```
> 
> @marco-c Isn't it supposed to be something like:
> 
> ```
> taskcluster_firefox_ci = { client_id = "<CLIENT_ID_NAME>", access_token = "<ACCESS_TOKEN>" }
> ```

Yes, the secrets are stored as yaml but the taskcluster API returns them as json.

EvaBardou:
> Yes, the secrets are stored as yaml but the taskcluster API returns them as json.

Ah ok, I thought you were using `.toml` files too to define the secret, that's what I'm using on my local instance

marco-c:
I tested this on push 2ef92e77b85b907bb7c8f570133ff77e8b32fbfd with task FxGCVpklQdK2Bbbmsp5Zxw. It correctly triggered the backfill task fzhDov9ZT8mHoboHik6jTQ, which in turn correctly triggered `test-linux1804-64-qr/opt-mochitest-plain-fis-e10s-2` (2-2ef92e77b85-bk) backfilled tasks in previous pushes.

marco-c (approved):
Thanks!

Issue #697, opened 2022-03-17
If a group is still running on a parent push, don't consider a regression as new


marco-c:
E.g. in case of backfills, there is a chance a group will finish on a child push before a parent push.
If there is a failure happening on a push, and the same group is still running on the parent, then we can't tell if the failure is new or pre-existing.

Not sure we can identify this situation though, as the task running the group is still running, and so errorsummary.log is not available yet.

marco-c:
For example, on 58fa4a284e789a372e56394703427e59baa34011 we had devtools/client/debugger/test/mochitest/browser.ini running and failing. On its parent push the task that was supposed to run the same group was retried, so it ended up running (and failing) after the equivalent task on the child push.
While the task on the parent was still running, the task on 58fa4a284e789a372e56394703427e59baa34011 was retriggered 5 times and failed 5 times, so we were considering devtools/client/debugger/test/mochitest/browser.ini as a regression of 58fa4a284e789a372e56394703427e59baa34011.

La0:
@marco-c In this case, how do we consider a group to be running:
- when any task of the group is not in a final state ?
- when we did not get all task results ?
- both ?

I *think* the second option is either to implement

marco-c:
There is one problem to figure out: right now until a task is complete, we don't know what groups it run!
We need to figure out if it's feasible to know which groups a task is going to run even before it runs or while it is still running.

Otherwise, we have to consider any group as potentially still running on any given push, unless it already run on that push or unless all tasks on that push are complete.

@ahal do you know if we can retrieve the groups that a task is going to run? Maybe from some artifact in the decision task? Or from the task definition?

marco-c:
We have MOZHARNESS_TEST_PATHS in the task definition that lists the groups that are going to run (this is only for some test suites, not all, so we'll need to have different paths of execution for different test suites).

We could add a `is_running` function on the Push that takes as input a group and:
1) iterates through all tasks to find those that are not complete;
2) if the group belongs to a test suite where MOZHARNESS_TEST_PATHS is not available, then we return True if there is any task still running for that test suite;
3) if the group belongs to a test suite where MOZHARNESS_TEST_PATHS is available, then we return True if there is any task still running where the group is in the MOZHARNESS_TEST_PATHS for that task.

marco-c:
> @marco-c A few questions:
> 
> > iterates through all tasks
> 
> So we need to iterate over all tasks on the `Push` using `self.tasks` right?

Yes.

[...]

PR #771, opened 2022-06-06
Extend the classify command to perform backfills and retriggers

### LINKED ISSUE #654: Extend the classify command to perform backfills (and choose between retriggers or backfills) when it needs more information
The algorithm behind the classify command isn't always able to give a decision between "good" or "bad" and sometimes needs more information in order to be sure.

When a situation like this happens and the classify command returns "unknown", it should automatically identify and perform the set of actions (retriggers and/or backfills) that would be necessary to gather enough information.

#619 implemented support for retriggers, but retriggers might not always be the best action. See https://github.com/mozilla/mozci/pull/619#issuecomment-1033590959.

Depends on https://github.com/mozilla/mozci/issues/504 and https://github.com/mozilla/mozci/issues/508 and https://github.com/mozilla/mozci/issues/743.

EvaBardou:
Closes #654 

@marco-c Please do not hesitate to let me know if there is anything I have misunderstood from the steps you specified in the issue.

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -244,24 +244,33 @@ class ClassifyCommand(Command):
                         retrigger(tasks=tasks, repeat_retrigger=1)
 
                 def retrigger_failures(groups, count):
-                    if not groups:
+                    groups_with_failures = {
+                        name: failing_tasks
+                        for name, failing_tasks in groups.items()
+                        if failing_tasks
+                    }
+                    if not groups_with_failures:
                         return
-                    # If there is more than one group failing, we should retrigger only one of them
-                    _, failing_tasks = next(iter(groups.items()))
+
+                    # If there is more than one group with failing tasks, we should retrigger only one of them
+                    _, failing_tasks = next(iter(groups_with_failures.items()))
                     retrigger(tasks=failing_tasks, repeat_retrigger=count)
 
                 retrigger_failures(
                     to_retrigger_or_backfill.real_retrigger,
-                    classify_parameters["consistent_failures_counts"][1],
+                    classify_parameters.get("consistent_failures_counts", (2, 3))[1],
                 )
                 retrigger_failures(
                     to_retrigger_or_backfill.intermittent_retrigger,
-                    classify_parameters["consistent_failures_counts"][0],
+                    classify_parameters.get("consistent_failures_counts", (2, 3))[0],
                 )
 
-                for index, (_, failing_tasks) in enumerate(
-                    to_retrigger_or_backfill.backfill.items()
-                ):
+                groups_to_backfill = {
+                    name: failing_tasks
+                    for name, failing_tasks in to_retrigger_or_backfill.backfill.items()
+                    if failing_tasks
+                }
+                for index, (_, failing_tasks) in enumerate(groups_to_backfill.items()):
                     # Only backfill X failing groups using --backfill-limit
                     if index == backfill_limit:
                         break
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -2,12 +2,13 @@
 import collections
 import csv
 import datetime
+import fnmatch
 import json
 import os
 import re
 import traceback
 from inspect import signature
-from typing import Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Set, Tuple
 from urllib.parse import urlencode
 
 import arrow
@@ -25,6 +26,7 @@ from mozci.push import (
     Push,
     PushStatus,
     Regressions,
+    ToRetriggerOrBackfill,
     make_push_objects,
     retrigger,
 )
@@ -229,6 +231,9 @@ class ClassifyCommand(Command):
                 "<comment>Provided --output pointed to a inexistent directory that is now created.</comment>"
             )
 
+        retriggerable_backfillable_patterns = config.get(
+            "retriggerable-backfillable-task-names", []
+        )
         try:
             backfill_limit = int(self.option("backfill-limit"))
         except ValueError:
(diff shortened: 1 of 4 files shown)


mozci/console/commands/push.py
```diff
@@ -225,12 +226,54 @@ def handle(self) -> None:
...
+
+                def retrigger_failures(groups, count):
+                    groups_with_failures = {
+                        name: failing_tasks
+                        for name, failing_tasks in groups.items()
+                        if failing_tasks
+                    }
+                    if not groups_with_failures:
+                        return
+
```
marco-c, line 251: Could you assert here that all failing tasks did fail?

mozci/console/commands/push.py
```diff
@@ -225,12 +226,54 @@ def handle(self) -> None:
...
 
+        try:
+            backfill_limit = int(self.option("backfill-limit"))
+        except ValueError:
+            raise Exception("Provided --backfill-limit should be an int.")
+
         for push in pushes:
             try:
-                classification, regressions = push.classify(**classify_parameters)
+                classification, regressions, to_retrigger_or_backfill = push.classify(
```
marco-c, line 236: There is a problem to solve: if we run the script and retrigger something, we don't want to retrigger again on the next run if the previously retriggered tasks haven't finished yet.
We should make sure not to retrigger if there is any task in progress that is running the same groups (similar logic as #697).

mozci/push.py
```diff
@@ -1207,12 +1223,35 @@ def _map_failing_tasks(groups):
             # Link all the failing tasks on the given groups
             return {name: self.group_summaries[name].failing_tasks for name in groups}
 
+        real_failures_to_be_retriggered = (groups_regressions & groups_high).difference(
```
marco-c, line 1226: Can you add comments explaining how we decide the groups to retrigger and backfill?
You can simply copy & paste my comments from the issue and reword them if necessary

marco-c:
We should add a config option to filter task names that we are allowed to retrigger/backfill (e.g. we might only want to retrigger/backfill "*-linux-*" tasks).

mozci/push.py
EvaBardou: I was writing the comments and wondered if it would be enough to link [your comment](https://github.com/mozilla/mozci/issues/654#issuecomment-1139488070) directly in the code to avoid repeating everything?

mozci/console/commands/push.py
marco-c: Yeah, let's do #697 first.

mozci/console/commands/push.py
marco-c: Not really necessary at the moment, but better to be defensive to potential future changes given that retriggers and backfills can be very expensive.

mozci/push.py
marco-c: Sounds good!

marco-c:
> > We should add a config option to filter task names that we are allowed to retrigger/backfill (e.g. we might only want to retrigger/backfill "_-linux-_" tasks).
> 
> @marco-c so we could imagine having a new parameter in the config called `retriggerable-backfillable-task-names` storing a list of patterns that we allow to retrigger/backfill like what we did for `autoclassification["test-suite-names"]`? If the parameter is empty do we allow to retrigger/backfill everything? Or nothing?

Sounds good. If the parameter is "*" we allow retriggering/backfilling everything, if the parameter is empty we allow nothing (so we have a way to totally disable it).

mozci/console/commands/push.py
marco-c: Or better, let's implement `is_group_running` first, we don't have to fix the full #697.

EvaBardou: Will work on it early next week hopefully. :crossed_fingers:

mozci/console/commands/push.py
```diff
@@ -787,8 +877,8 @@ def handle(self) -> None:
                     )
 
             if push.backedout or push.bustage_fixed_by:
-                group_classifications: dict[
-                    str, dict[tuple[str, str], set[str]]
+                group_classifications: Dict[
+                    str, Dict[Tuple[str, str], Set[str]]
```
marco-c, line 881: Nit: why did you move to typing instead of built-in names?

mozci/console/commands/push.py
```diff
@@ -314,6 +335,75 @@ def _serialize_regressions(regressions):
...
+            allowed_patterns,
+        )
+
+        # Backfill some failures
+        groups_to_backfill = {
+            name: failing_tasks
+            for name, failing_tasks in to_retrigger_or_backfill.backfill.items()
+            if failing_tasks
+        }
+        for index, (_, failing_tasks) in enumerate(groups_to_backfill.items()):
```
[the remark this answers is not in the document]
marco-c: Or something like this should work too, so you don't have to manually implement the limit:
```Python
for failing_tasks in itertools.islice(groups_to_backfill.values(), backfill_limit):
```

mozci/console/commands/push.py
```diff
@@ -314,6 +335,75 @@ def _serialize_regressions(regressions):
...
+    def retrigger_failures(self, groups, count, allowed_patterns):
+        groups_with_failures = {
+            name: failing_tasks
+            for name, failing_tasks in groups.items()
+            if failing_tasks
+        }
+        if not groups_with_failures:
+            return
+
+        # If there is more than one group with failing tasks, we should retrigger only one of them
```
marco-c, line 347: Sorry I realize this wasn't clear in https://github.com/mozilla/mozci/issues/654#issuecomment-1145889375.
We should retrigger all groups, and if there is more than one task for a given group we should retrigger only one of the tasks.

mozci/console/commands/push.py
```diff
@@ -207,6 +209,7 @@ class ClassifyCommand(Command):
         {--show-intermittents : If set, print tasks that should be marked as intermittent.}
         {--environment=testing : Environment in which the analysis is running (testing, production, ...)}
         {--retrigger-unknown : If set, retrigger UNKNOWN regressions}
+        {--backfill-limit=3 : Number of failing groups (missing information) to be backfilled.}
```
marco-c, line 212: Let's also add a --retrigger-limit.

mozci/console/commands/push.py
EvaBardou: Linting was failing:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

mozci/console/commands/push.py:880: error: "dict" is not subscriptable, use "typing.Dict" instead
mozci/console/commands/push.py:880: error: "tuple" is not subscriptable, use "typing.Tuple" instead
mozci/console/commands/push.py:880: error: "set" is not subscriptable, use "typing.Set" instead
Found 3 errors in 1 file (checked 39 source files)
```

The change was introduced by https://github.com/mozilla/mozci/pull/781 and merged with a failing pipeline

marco-c: Oh no, I did notice, it was fixed in #784.

EvaBardou: So it might be because I'm using python 3.8, do you want me to revert the change?

EvaBardou: Python 3.10 doesn't like me ahah:
> The virtual environment found in /home/eva/.cache/pypoetry/virtualenvs/mozci-BCG4KwUI-py3.10 seems to be broken.

Also you are using typing everywhere in mozci not built-in names

marco-c: Yeah we were using typing because the feature to use built-in names was not available yet, but it is the future

mozci/console/commands/push.py
```diff
@@ -314,6 +339,76 @@ def _serialize_regressions(regressions):
...
+            if failing_tasks
+        }
+        for failing_tasks in itertools.islice(
+            groups_to_backfill.values(), 0, backfill_limit
+        ):
+            for t in failing_tasks:
+                if t.label and any(
+                    fnmatch.fnmatch(t.label, pattern) for pattern in allowed_patterns
+                ):
+                    t.backfill(push)
```
marco-c, line 410: @Archaeopteryx should we notify the group on Matrix when we trigger a backfill?

Archaeopteryx: No, failures in the backfills is what is important.

PR #778, opened 2022-06-07
Backfill with times set to 5 when a task is not classified

### LINKED ISSUE #776: When backfilling, use `5` as times when there are no classifications
Currently we use 5 only when the task was classified as intermittent, but at the time we backfill we don't know yet if the task is intermittent as it might not have been classified yet.

marco-c:
Fixes #776

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -302,7 +302,9 @@ class Task:
                 "taskId": self.id,
                 "taskGroupId": decision_task.id,
                 "input": {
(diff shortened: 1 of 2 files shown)


EvaBardou:
@marco-c LGTM but tests are failing and should be updated, it also might be a good idea to add a new case to the `@pytest.mark.parametrize` of `test_backfill` to assert that a classification differing from `not classified` or `intermittent` will only send `times=1`

EvaBardou (approved):

PR #779, opened 2022-06-09
Warn if there is a classification that references a revision that does not exist

marco-c:
Linked to #739

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -18,7 +18,7 @@ from tabulate import tabulate
 from taskcluster.exceptions import TaskclusterRestFailure
 
 from mozci import config
-from mozci.errors import SourcesNotFound, TaskNotFound
+from mozci.errors import PushNotFound, SourcesNotFound, TaskNotFound
 from mozci.push import (
     MAX_DEPTH,
     Push,
@@ -29,6 +29,7 @@ from mozci.push import (
 )
 from mozci.task import Task, TestTask, is_autoclassifiable
 from mozci.util.defs import INTERMITTENT_CLASSES
+from mozci.util.hgmo import HgRev
 from mozci.util.taskcluster import (
     COMMUNITY_TASKCLUSTER_ROOT_URL,
     get_taskcluster_options,
@@ -586,6 +587,20 @@ class ClassifyEvalCommand(Command):
                 self.line(
                     f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
                 )
+
+            for task in p.tasks:
+                for classification, note in task.classifications:
+                    if classification != "fixed by commit":
+                        continue
+
+                    try:
+                        fix_hgmo = HgRev.create(note[:12], branch=self.branch)
+                        if len(fix_hgmo.backouts) == 0:
+                            continue
+                    except PushNotFound:
+                        self.line(
+                            f"<comment>Task {task.id} on push {push.branch}/{push.rev} contains a classification that references a non-existent revision: {note}</comment>"
+                        )
 
             # Advance the overall progress bar
             progress.advance()
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -588,7 +588,7 @@ class ClassifyEvalCommand(Command):
                     f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
                 )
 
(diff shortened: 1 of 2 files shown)


mozci/console/commands/push.py
```diff
@@ -587,6 +588,20 @@ def handle(self) -> None:
                     f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
                 )
 
+            for task in p.tasks:
```
EvaBardou, line 591:
```suggestion
            for task in push.tasks:
```

Otherwise this line will raise an error:
```
  UnboundLocalError

  local variable 'p' referenced before assignment
```

mozci/console/commands/push.py
```diff
@@ -587,6 +588,20 @@ def handle(self) -> None:
...
+                for classification, note in task.classifications:
```
EvaBardou, line 592: I think only `GroupSummary` objects have the `classifications` property

```
  AttributeError

  'Task' object has no attribute 'classifications'
```

mozci/console/commands/push.py
```diff
@@ -587,6 +588,20 @@ def handle(self) -> None:
...
+                    if classification != "fixed by commit":
+                        continue
+
+                    try:
+                        fix_hgmo = HgRev.create(note[:12], branch=self.branch)
+                        if len(fix_hgmo.backouts) == 0:
```
EvaBardou, line 598: What's the use for this line since you'll continue anyway? Would it possibly raise another `PushNotFound` exception needed to trigger the warning?

mozci/console/commands/push.py
```diff
@@ -587,6 +588,20 @@ def handle(self) -> None:
                     f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
                 )
 
+            for task in p.tasks:
+                for classification, note in task.classifications:
+                    if classification != "fixed by commit":
+                        continue
+
+                    try:
+                        fix_hgmo = HgRev.create(note[:12], branch=self.branch)
```
EvaBardou, line 597:
```suggestion
                        fix_hgmo = HgRev.create(note[:12], branch=branch)
```

EvaBardou (changes requested):

mozci/console/commands/push.py
marco-c: Oops, I copied and pasted from a separate script that I wrote without then actually testing the command :P
Relying too much on linting to catch these errors!

mozci/console/commands/push.py
marco-c: Lazy copy and paste. I'll remove it.
We could in theory check if the referenced revision contains either a backout of a bustage fix, but bustage fixes are hard to detect so the check wouldn't be precise.

mozci/console/commands/push.py
marco-c: Fixed in c071b96.

mozci/console/commands/push.py
marco-c: Fixed in 27446a1.

mozci/console/commands/push.py
marco-c: Fixed in a9b467a.

mozci/console/commands/push.py
```diff
@@ -587,6 +588,17 @@ def handle(self) -> None:
                     f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
                 )
 
+            for task in push.tasks:
+                if task.classification != "fixed by commit":
+                    continue
+
+                try:
+                    HgRev.create(task.classification_note[:12], branch=push.branch)
```
EvaBardou, line 596: I tried to modify a task classification by putting nonsense and the warning did not appear, it doesn't look like `HgRev.create()` will ever raise a `PushNotFound` exception.

```
task.classification = "fixed by commit"
task.classification_note = "not a revision"
```

EvaBardou (changes requested):

mozci/console/commands/push.py
marco-c: Fixed in 069e84b1b6fb209971591f8529d921436eac1396.
"not a revision" will actually cause another exception to be thrown, different than PushNotFound. I'll fix that in another commit.

marco-c: Done in d5764e0eef2185366c29f39f38e82935882cb6e5.

EvaBardou (approved):
LGTM, thanks!

PR #782, opened 2022-06-11
A couple of minor improvements to the warning about backedout pushes that are never classified as cause of any failure

marco-c:
Linked to #739


marco-c:
Select to hide whitespaces when reviewing this, it's much easier :)

marco-c:
It's actually a trivial patch, I'll just merge it.

PR #791, opened 2022-06-20
Filter out groups that are still running from real failures

### LINKED ISSUE #697: If a group is still running on a parent push, don't consider a regression as new

EvaBardou:
Closes #697

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -1253,7 +1253,7 @@ class Push:
                     break
 
         logger.debug(
-            f"Got {len(groups_still_running)} groups still running in the push"
+            f"Got {len(groups_still_running)} groups failing in this push but still running in a parent push, so we can't know if they are regressions from this push"
         )
         real_failures -= groups_still_running
         unknown_failures |= groups_still_running


EvaBardou:
@marco-c I'm not sure how we can test this though, if you have any idea, don't hesitate :smile:

mozci/push.py
```diff
@@ -533,6 +535,39 @@ def scheduled_duration(self):
...
+        """Checks if the provided group is still running on this push.
+
+        Returns:
+            bool: True, if the group is still running.
+                  False, if the group is completed.
+        """
+        running_tasks = [
+            task
+            for task in self.tasks
+            if task.state not in ["completed", "failed", "exception"]
```
marco-c, line 548: This makes me think that we should update `completed_cached_tasks` to also consider `failed` and `exception`, we were only caching `completed` :)

mozci/push.py
```diff
@@ -533,6 +535,39 @@ def scheduled_duration(self):
...
+                ):
+                    return True
+        else:
+            group_types = set([task.tags.get("tests-type", "") for task in group.tasks])
+            running_types = set(
+                [task.tags.get("tests-type", "") for task in running_tasks]
+            )
+            return any(
+                tests_type in running_types for tests_type in group_types if tests_type
+            )
```
marco-c, line 567: Nit: this could be simplified by doing something like:
```Python
return not group_types.isdisjoint(running_types)
```
or
```Python
return len(group_types | running_types) != 0
```

mozci/push.py
```diff
@@ -533,6 +535,39 @@ def scheduled_duration(self):
...
+
+        return False
```
marco-c, line 569: Nit: move this in the first `if` branch after the for, as the else branch always returns

mozci/push.py
```diff
@@ -1181,10 +1219,15 @@ def classify_regressions(
...
+        )
 
-        # Real failure are groups with likely regressions that were selected with high confidence
-        # AND failing across config
-        real_failures = groups_regressions & groups_relevant_failure & groups_high
+        # Real failures are groups that finished running, with likely regressions that were selected
+        # with high confidence AND failing across config
+        real_failures = (
+            groups_regressions & groups_relevant_failure & groups_high
+        ) - groups_still_running
```
marco-c, line 1230: We should move these still running groups to the unknown failures list (but take into account that we are not going to do any actions on them as part of #654.

mozci/task.py
```diff
@@ -220,6 +220,10 @@ def is_backfill(self) -> bool:
     def is_retrigger(self) -> bool:
         return self.tags.get("action", "").startswith("retrigger-")
 
+    @property
+    def is_tests_grouped(self) -> bool:
```
marco-c, line 224: Nit: could you use this function in the `label_summaries` function in push.py too?

mozci/task.py
```diff
@@ -303,7 +307,8 @@ def backfill(self, push):
                 "taskGroupId": decision_task.id,
                 "input": {
                     "times": 5
-                    if self.classification == "not classified" or self.classification in INTERMITTENT_CLASSES
+                    if self.classification == "not classified"
+                    or self.classification in INTERMITTENT_CLASSES
```
marco-c, line 311: Nit: could you revert this unnecessary change?

marco-c:
> @marco-c I'm not sure how we can test this though, if you have any idea, don't hesitate smile

You could ask @Archaeopteryx via chat to tell you when sheriffs backfill a task for reftest and one for mochitest, then we can try running the `is_group_running` function for one of the groups of that task and check that it is returning True until the backfill finishes (and False since then).

mozci/push.py
EvaBardou: Would you like me to make the update in this PR or elsewhere?

marco-c: You could do it in this PR, defining this list in the defs.py file

mozci/task.py
marco-c: Weird that it doesn't fail on master :S
Anyway, I've run it locally and pushed the reformatted code to master, so you can rebase.

mozci/push.py
```diff
@@ -533,6 +535,44 @@ def scheduled_duration(self):
...
+                    return True
+            return False
+
+        group_types = set(
+            [
+                task.tags["tests-type"]
+                for task in group.tasks
+                if task.tags.get("tests-type")
+            ]
+        )
```
marco-c, line 566: Nit: no need to create a list first:
```Python
        group_types = set(
                task.tags["tests-type"]
                for task in group.tasks
                if task.tags.get("tests-type")
        )
```
or
```Python
        group_types = {
                task.tags["tests-type"]
                for task in group.tasks
                if task.tags.get("tests-type")
        }
```

mozci/push.py
```diff
@@ -533,6 +535,44 @@ def scheduled_duration(self):
...
+        running_types = set(
+            [
+                task.tags["tests-type"]
+                for task in running_tasks
+                if task.tags.get("tests-type")
+            ]
+        )
```
marco-c, line 573: Nit: same here, no need to create a list first

EvaBardou:
@marco-c A couple of tests are failing in CI due to this change https://github.com/mozilla/mozci/pull/791/files#diff-cbdcefec12b605b952245761dbfbbbb2ec7fc73addcc510c2b8dc48a60bd5d02R345, I can reproduce locally, it is altering the behavior of the `classify` function, what should we do about this?

mozci/push.py
```diff
@@ -1169,6 +1205,9 @@ def classify_regressions(
             for g in all_groups
             if g.name not in list(bugbug_selection["groups"].keys())
         }
+
+        groups_still_running = {g.name for g in all_groups if self.is_group_running(g)}
```
marco-c, line 1209: We should actually check if the group is still running in any of the parents of the push.
We iterate the parents and we either stop if we find the group run, or add the group to the groups_still_running set.
Pseudocode:
```Python
for group in real_failures:
    for parent in push.parents:
        # If the group run in a parent, then we don't care if it's still running in a grandparent.
        if group in parent.group_summaries:
            break

        if parent.is_group_running(group):
            groups_still_running.add(group)
            break

real_failures -= groups_still_running
unknown_failures |= groups_still_running
```

EvaBardou:
> Are the tests still failing after my latest changes?

Unfortunately, yes

> If yes, we should make sure in the mocks we are not considering any group as still running, then the tests should behave exactly the same.

We aren't, I patched the function to always return `False` with: `monkeypatch.setattr(Push, "is_group_running", lambda *args: False)`
But if a group isn't running, currently it is added to `unknown_failures`, thus changing the classified failures and the overall classification of the Push in some cases.

marco-c:
Let's see if it still happens after https://github.com/mozilla/mozci/pull/791#discussion_r902901661, but it shouldn't happen if you mock all groups to consider them as finished

PR #804, opened 2022-07-01
Notify the Matrix room for all warnings from #739

### LINKED ISSUE #780: Notify Matrix channel on the warnings from #739
Right now we just add the warnings to the email, but we should also send a message to Matrix so sheriffs can take countermeasures.

EvaBardou:
Closes #780 

@marco-c You might need to add the `matrix-room-id` attribute in the `classify-eval` command secret, I don't remember if the secret is shared between all commands or if we use separate ones...

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -324,7 +324,7 @@ class ClassifyCommand(Command):
 
             # Send a notification when some emails are declared in the config
             emails = config.get("emails", {}).get("classifications")
-            matrix_room = config.get("matrix-room-id", None)
+            matrix_room = config.get("matrix-room-id")
             if emails or matrix_room:
                 # Load previous classification from taskcluster
                 try:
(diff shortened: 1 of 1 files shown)


mozci/console/commands/push.py
```diff
@@ -792,75 +812,53 @@ def handle(self) -> None:
...
+                        push, "group"
+                    )
 
                 if not ever_classified_as_cause:
-                    self.line(
-                        f"<comment>Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push</comment>"
+                    warnings.append(
+                        {
+                            "message": f"Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push.",
+                            "type": "comment",
```
marco-c, line 829:
```suggestion
                            "type": "error",
```

mozci/console/commands/push.py
```diff
@@ -792,75 +812,53 @@ def handle(self) -> None:
...
                 )
                 try:
                     fix_hgmo.changesets
                 except PushNotFound:
-                    self.line(
-                        f"<comment>Task {task.id} on push {push.branch}/{push.rev} contains a classification that references a non-existent revision: {task.classification_note}</comment>"
+                    warnings.append(
+                        {
+                            "message": f"Task {task.id} on push {push.branch}/{push.rev} contains a classification that references a non-existent revision: {task.classification_note}.",
+                            "type": "comment",
```
marco-c, line 847:
```suggestion
                            "type": "error",
```

marco-c (approved):
Looks good, thanks!

PR #809, opened 2022-07-07
Add config options to easily enable/disable warnings

marco-c:


mozci/console/commands/push.py
```diff
@@ -916,14 +925,16 @@ def handle(self) -> None:
...
+                            warnings_to_notify.add(warnings[-1]["message"])
 
             # Output all warnings and also send them to the Matrix room if defined
             matrix_room = config.get("matrix-room-id")
             for warning in warnings:
                 warn_type = warning["type"]
                 warn_message = warning["message"]
                 self.line(f"<{warn_type}>{warn_message}</{warn_type}>")
-                if matrix_room:
+                if matrix_room and warn_message in warnings_to_notify:
```
EvaBardou, line 937: Instead of using a new set maybe we could just add a new boolean attribute `to_notify` in each warning dict from the `warnings` list, WDYT?

mozci/console/commands/push.py
```diff
@@ -822,13 +823,15 @@ def handle(self) -> None:
                         push, "group"
                     )
 
-                if not ever_classified_as_cause:
```
EvaBardou, line 825: This line shouldn't be removed

mozci/console/commands/push.py
```diff
@@ -822,13 +823,15 @@ def handle(self) -> None:
...
-                    warnings.append(
-                        {
-                            "message": f"Push {push.branch}/{push.rev} was backedout and all of its failures and the ones of its children were marked as intermittent or marked as caused by another push.",
-                            "type": "error",
-                        }
-                    )
+                    if config.get("warnings", {}).get(
```
EvaBardou, line 826: You're not using `warnings_to_notify` here like for the other warnings

EvaBardou (dismissed):

mozci/console/commands/push.py
marco-c: Good idea, done in ed9738f.

mozci/console/commands/push.py
marco-c: Oops, fixed in 7326dfc.

mozci/console/commands/push.py
marco-c: Fixed by dropping `warnings_to_notify` in ed9738f.

Issue #832, opened 2022-08-25
Take into account failure message for autoclassifications


marco-c:
E.g. if a failure message has already been seen for a known intermittent, we know it's more likely to be intermittent. If a failure message is new, we know it's more likely not to be an intermittent.

marco-c:
We can now rely on the new classification type @jmaher has added in https://github.com/mozilla/treeherder/pull/7691.

PR #970, opened 2023-05-16
Add support for 'new failure not classified'

jmaher:


mozci/console/commands/push.py
```diff
@@ -611,6 +611,8 @@ def parse_and_log_details(
             continue
         if classifications_set == {"not classified"}:
```
marco-c, line 612: Here we want to check if classification_set only includes `"not classified"` and `"new failure not classified"`, so we should do something like:
```python
if len(classifications_set - {"not classified", "new failure not classified"}) == 0:
```
and then you can remove the `elif` you added.

mozci/task.py
```diff
@@ -346,7 +346,8 @@ def backfill(self, push):
                 "taskGroupId": decision_task.id,
                 "input": {
                     "times": 5
-                    if self.classification == "not classified"
+                    if self.classification
+                    in ["not classified", "new failure not classified"]
```
marco-c, line 350: Nit:
```suggestion
                    in ("not classified", "new failure not classified")
```
(tuple is immutable, so will consume slightly less memory)

marco-c (commented):
Looks good overall, just a minor problem to fix and a nit.

We should also file a follow-up issue about using the information from NEW, I will do that tomorrow.

marco-c (approved):

marco-c:
Turns out we already have an issue that serves the purpose: #832.

PR #1045, opened 2023-10-10
Consider groups classified as "NEW" to be potentially regressions and groups not classified as "NEW" to be potentially intermittents

### LINKED ISSUE #832: Take into account failure message for autoclassifications

marco-c:
Fixes #832

In practice, this means a group which is not classified as "NEW" and which is failing on a config and passing on another or failing at first and passing on a retrigger, will be autoclassified as intermittent.

Depends on #1044.

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -1306,8 +1306,10 @@ class Push:
         # See the following comment that explains how we decide the groups to retrigger/backfill
         # https://github.com/mozilla/mozci/issues/654#issuecomment-1139488070
         real_failures_to_be_retriggered = (
-            groups_regressions & groups_high.union(groups_classified_as_new)
-        ) - groups_consistent_failure
+            (groups_regressions & groups_high.union(groups_classified_as_new))
+            - groups_consistent_failure
+            - groups_non_consistent_failure
+        )
         real_groups_still_running = groups_still_running | {
             group_name
             for group_name in real_failures_to_be_retriggered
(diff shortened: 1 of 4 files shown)


mozci/push.py
```diff
@@ -1302,8 +1317,8 @@ def _map_failing_tasks(groups):
 
         intermittent_failures_to_be_retriggered = (
             set(g.name for g in push_groups if g.status != Status.PASS)
-            & groups_low.union(groups_no_confidence)
-        ) - groups_non_relevant_failure
+            & groups_low.union(groups_no_confidence).union()
+        ) - groups_non_consistent_failure
         intermittent_groups_still_running = groups_still_running | {
             group_name
             for group_name in intermittent_failures_to_be_retriggered
```
jmaher, line 1324: this will take some logic to figure out- in here we look for is running, but if it is a retrigger or a confirm-failure, it will need to account for both.

jmaher (approved):

mozci/push.py
marco-c: It is using `MOZHARNESS_TEST_PATHS` to figure out which groups (i.e. manifests) are running in a task. Is `MOZHARNESS_TEST_PATHS` set on confirm-failure tasks?

jmaher: yes, it is the full path, not the manifest name, i.e. `"MOZHARNESS_TEST_PATHS": "{\"mochitest-plain\": [\"dom/html/test/test_bug1166138.html\"]}"`

marco-c: Oh, we will have to adjust this either in mozci or in mozilla-central. In other tasks, MOZHARNESS_TEST_PATHS contains manifest names.
Could we change confirm-failures to store manifests in MOZHARNESS_TEST_PATHS like other tasks do and the test paths in another env variable?

jmaher: I would need a method for determining which test file to run since I am not running a manifest but a single test.  For test-verify, we have a patch with changed files and we look for matching tests paths in those changed files;  I am not thinking of a quick and easy way to solve this in confirm-failures; possibly another env variable.

jmaher: I have patches up for review to leave `MOZHARNESS_TEST_PATHS` alone

PR #1053, opened 2023-10-24
add logic to mozci for using and triggering confirm failures

jmaher:
this is to show a WIP

I still need to:
 * create unittests
 * remove my hack for 2 days (needed that for a try push <8 hours old)
 * decide on logic for how to use cf tasks

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -438,7 +438,7 @@ class ClassifyCommand(Command):
             groups_with_failures.values(), 0, retrigger_limit
         ):
             # If there is more than one task failing in this group, we should retrigger only one of them
-            failing_tasks[0].retrigger(push, count)
+            failing_tasks[0].confirm(push)
 
     def backfill_and_retrigger_failures(
         self,
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -1073,6 +1073,14 @@ class Push:
             if failure_summary.is_intermittent:
                 count *= 2
 
+            # if confirmed = True; then assume regression
+            # if confirmed = False; then assume it's very unlikely a regression
+            if isinstance(failure_summary, GroupSummary):
+                if failure_summary.is_confirmed_failure():
+                    count = 0
+                elif failure_summary.is_confirmed_failure() is False:
+                    count = MAX_DEPTH + 1
+
             if not prior_regression and count <= MAX_DEPTH:
                 regressions[name] = int(count) if count > 0 else 0
 
@@ -1189,28 +1197,37 @@ class Push:
         else:
             all_groups = push_groups
 
+        groups_confirmed = {g.name: g.is_confirmed_failure() for g in all_groups}
+
         groups_consistent_failure = {
             g.name
             for g in all_groups
-            if g.is_consistent_failure(
-                cross_config_counts[1] if cross_config_counts is not None else None,
-                consistent_failures_counts[1]
-                if consistent_failures_counts is not None
-                else None,
+            if groups_confirmed[g.name]
+            or (
+                groups_confirmed[g.name] is None
+                and g.is_consistent_failure(
+                    cross_config_counts[1] if cross_config_counts is not None else None,
+                    consistent_failures_counts[1]
+                    if consistent_failures_counts is not None
+                    else None,
+                )
             )
         }
 
         groups_non_consistent_failure = {
             g.name
             for g in all_groups
-            if g.status != Status.PASS
-            and g.is_consistent_failure(
-                cross_config_counts[0] if cross_config_counts is not None else None,
-                consistent_failures_counts[0]
-                if consistent_failures_counts is not None
-                else None,
+            if groups_confirmed[g.name] is False
+            or (
+                groups_confirmed[g.name] is None
+                and g.is_consistent_failure(
+                    cross_config_counts[0] if cross_config_counts is not None else None,
+                    consistent_failures_counts[0]
+                    if consistent_failures_counts is not None
+                    else None,
+                )
+                is False
             )
-            is False
         }
 
         groups_failing = {g.name for g in all_groups if g.status != Status.PASS}
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -357,6 +357,33 @@ class Task:
         logger.info("Retriggering task '{}'".format(self.tags.get("label", "")))
         return self._trigger_action(retrigger_action, hook_payload)
 
+    def confirm(self, push):
+        """This function implements ability to perform retriggers on tasks"""
+        if self._should_retrigger() == "false":
+            logger.info(
+                "Not confirming task '{}' failures, task should not be confirmed".format(
+                    self.tags.get("label")
+                )
+            )
+            return None
+
+        decision_task = push.decision_task
+        retrigger_action = self._get_action(decision_task, "confirm-failures")
+
+        hook_payload = jsone.render(
+            retrigger_action["hookPayload"],
+            context={
+                "taskId": self.id,
+                "taskGroupId": decision_task.id,
+                "input": {},
+            },
+        )
+
+        logger.info(
+            "confirming failures for task '{}'".format(self.tags.get("label", ""))
+        )
+        return self._trigger_action(retrigger_action, hook_payload)
+
     def _should_retrigger(self):
         """Return whether this task should be retriggered."""
         return self.tags.get("retrigger", "false")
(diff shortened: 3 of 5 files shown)


mozci/task.py
```diff
@@ -616,11 +643,28 @@ def failing_tasks(self):
 
     def is_config_consistent_failure(self, minimum_count: int = 3) -> Optional[bool]:
```
marco-c, line 644: I think we should keep the `is_config_consistent_failure` logic unchanged, and make sure the "-cf" results are considered as part of the runs anyway from mozci.

mozci/console/commands/push.py
```diff
@@ -438,7 +438,7 @@ def retrigger_failures(
             groups_with_failures.values(), 0, retrigger_limit
         ):
             # If there is more than one task failing in this group, we should retrigger only one of them
-            failing_tasks[0].retrigger(push, count)
+            failing_tasks[0].confirm(push)
```
marco-c, line 441: Should we `confirm` in cases it is supported and fall back on `retrigger`?

marco-c:
The change looks good to me overall

mozci/task.py
jmaher: where would you recommend changing the logic for `-cf` tasks? some things to consider:
1) CF will have a higher weight to confirm failure or prove intermittent
2) will need to match CF to task with same name without `-cf` suffix

marco-c: @jmaher regarding 1, it can be a follow-up. It could go in https://github.com/mozilla/mozci/blob/d8e8c8c4437124b9e6282aece221de0b6385529f/mozci/task.py#L496 or it could go in https://github.com/mozilla/mozci/blob/d8e8c8c4437124b9e6282aece221de0b6385529f/mozci/push.py#L1121.

Regarding 2, I think we just need to update https://github.com/mozilla/mozci/blob/d8e8c8c4437124b9e6282aece221de0b6385529f/mozci/task.py#L89 so that the result is the same for a task with or without -cf.

mozci/console/commands/push.py
jmaher: confirm will call retrigger; although the `count` is missing and needs to be dealt with

mozci/task.py
jmaher: thanks, I can move these;  I need to make sure when stripping `-cf` that we look at what specific test/group was run, so we don't adjust counts incorrectly.

jmaher: I have a whole new approach for this, basically I have a method on tasks `is_confirmed_failure()` -> `True`, `False`, `None`

here we can add some logic here utilizing this in other commands- I have an example.  Possibly this could be optimized a bit for clarity or cpu performance.

jmaher:
still remaining is unittests

mozci/push.py
```diff
@@ -1190,10 +1190,13 @@ def classify_regressions(
         else:
             all_groups = push_groups
 
+        groups_confirmed = {g.name: g.is_confirmed_failure() for g in all_groups}
```
jmaher, line 1193: note: I had an issue finding tasks when using this below similar to `g.is_consistent_failure(...)`, so I went this route where I got the status for each group name and then verified that name.

marco-c: Do you remember what the issue was? In theory it should just work the same way :/

mozci/push.py
```diff
@@ -1204,14 +1207,17 @@ def classify_regressions(
         groups_non_consistent_failure = {
             g.name
             for g in all_groups
-            if g.status != Status.PASS
-            and g.is_consistent_failure(
-                cross_config_counts[0] if cross_config_counts is not None else None,
-                consistent_failures_counts[0]
-                if consistent_failures_counts is not None
-                else None,
+            if groups_confirmed[g.name] is None
```
marco-c, line 1210: I would say both if `is None` and `is False` here, you could do 
```suggestion
            if not groups_confirmed[g.name]
```

mozci/task.py
```diff
@@ -614,6 +647,31 @@ def failing_tasks(self):
             )
         ]
 
+    def is_confirmed_failure(self) -> Optional[bool]:
+        # match labels to -cf labels, and if ran, return True || False
+        # if not run -cf task, return None
+        # return True = intermittent
+        # return False = regression
```
marco-c, line 654: I'd do the opposite `True` means it is a confirmed failure (so regression), `False` means it is not (so intermittent).
This would match better the name of the function (and its usage in mozci/push.py).

mozci/task.py
```diff
@@ -614,6 +647,31 @@ def failing_tasks(self):
...
+                if (
+                    task.label
+                    and task.label.endswith("-cf")
+                    and result.group == self.name
+                ):
+                    cfname = get_configuration_from_label(task.label.strip("-cf"))
+                    cf_to_results[cfname].append(result.ok)
+
+        for cfname in cf_to_results:
+            if cf_to_results[cfname]:
```
marco-c, line 668: nit: The if is unnecessary, because you are already iterating through keys in cf_to_results

mozci/push.py
jmaher: the issue was not all groups were shown, specifically groups from a -cf task.  I cannot remember what specifically what the root cause; I think there was a filter of some kind taking place earlier.

mozci/task.py
```diff
@@ -34,6 +34,7 @@ class Status(Enum):
     INTERMITTENT = 2
 
 
+# TODO: we should ingest an artifact from m-c with a list of suites, this is out of date already
```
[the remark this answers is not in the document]
jmaher: https://github.com/mozilla/mozci/pull/1071

mozci/push.py
jmaher: end result is similar to this, but not exactly.

mozci/push.py
```diff
@@ -1189,10 +1196,13 @@ def classify_regressions(
         else:
             all_groups = push_groups
 
+        groups_confirmed = {g.name: g.is_confirmed_failure() for g in all_groups}
+
         groups_consistent_failure = {
             g.name
             for g in all_groups
-            if g.is_consistent_failure(
+            if groups_confirmed[g.name] is True
```
marco-c, line 1204: nit:
```suggestion
            if groups_confirmed[g.name]
```
or:
```suggestion
            if g.name in groups_confirmed
```

mozci/push.py
```diff
@@ -1076,6 +1076,13 @@ def get_regressions(
             if not prior_regression and count <= MAX_DEPTH:
                 regressions[name] = int(count) if count > 0 else 0
 
+            # if confirmed = True; then assume regression
```
marco-c, line 1079:
```suggestion
            # if confirmed = True; then assume regression
            # if confirmed = False; then assume it's very unlikely a regression
```

mozci/push.py
```diff
@@ -1076,6 +1076,13 @@ def get_regressions(
...
+            if isinstance(failure_summary, GroupSummary):
+                if failure_summary.is_confirmed_failure():
+                    count = 0
+                elif failure_summary.is_confirmed_failure() is False:
+                    count += 1
```
marco-c, line 1084: Maybe:
```suggestion
                    count *= 2
```
or even `count = MAX_DEPTH + 1` so it is removed from the list of regressions?

marco-c, line 1084: This code should be moved before `count` is used a few lines above `if not prior_regression and count <= MAX_DEPTH:`.

mozci/push.py
```diff
@@ -1189,10 +1196,13 @@ def classify_regressions(
...
+
         groups_consistent_failure = {
             g.name
             for g in all_groups
-            if g.is_consistent_failure(
+            if groups_confirmed[g.name] is True
+            or g.is_consistent_failure(
                 cross_config_counts[1] if cross_config_counts is not None else None,
                 consistent_failures_counts[1]
                 if consistent_failures_counts is not None
```
marco-c, line 1208: With this approach, if a manifest is failing multiple times but the corresponding -cf tasks are not failing (so g.is_confirmed is False and g.is_consistent_failure is True) we still consider a group as a real failure. Should we instead do something like:
```g.is_confirmed is True or (g.is_confirmed is None and g.is_consistent_failure)```?

File: mozci/push.py [hunk 12]
```diff
@@ -1203,14 +1213,17 @@ def classify_regressions(
...
+                and g.is_consistent_failure(
+                    cross_config_counts[0] if cross_config_counts is not None else None,
+                    consistent_failures_counts[0]
+                    if consistent_failures_counts is not None
+                    else None,
+                )
+                is False
             )
-            is False
         }
```
marco-c, line 1227: Similarly here, with this approach, if a manifest is not failing multiple times but the corresponding -cf tasks are failing (so g.is_confirmed is True and g.is_consistent_failure is False) we still consider a group as an intermittent failure. Should we instead do something like:
```g.is_confirmed is False or (g.is_confirmed is None and g.is_consistent_failure is False)```?

mozci/push.py
jmaher: yes, that makes sense;  this helps rule out the edge case where confirm-failure fails for reasons of test issues but the failure occurs in the scope of the manifest consistently.

This does bring up a different edge case, specifically:
 * task1: groupA: test1 fails
 * task2: groupA: test2 fails
 * task3: groupA: test3 fails
 * cf-task1: groupA: test1 passes
 
 this will look like groupA fails ALL the time but not in CF, so we would fail here.

mozci/push.py
jmaher: good catch, otherwise here we would end up allowing `g.is_confirmed and !g.is_consistent_failure`

jmaher:
thanks for the review @marco-c , I have updated accordingly, please take a few minutes to review so I can address further issues or merge it in :)

marco-c (approved):
Thanks!

PR #1071, opened 2023-11-29
dynamically load suites

jmaher:


mozci/task.py
```diff
@@ -34,46 +35,37 @@ class Status(Enum):
...
+    SUITES: set[str] = set()
+
+    def __init__(self):
+        # code used to populate SUITE
+        try:
+            r = requests.get(
+                "https://hg.mozilla.org/mozilla-central/raw-file/tip/taskcluster/ci/test/test-sets.yml"
+            )
+            y = yaml.load(r.text, yaml.BaseLoader)
+        except requests.ConnectionError:
```
marco-c, line 48: I would not catch the exception and let if fail, so the user knows something is wrong. If we fail silently, it could cause random problems which are difficult to debug

mozci/task.py
```diff
@@ -34,46 +35,37 @@ class Status(Enum):
...
+            y = {}
+            pass
+        test_types = set()
+        for item in y:
+            test_types |= set(y[item])
```
marco-c, line 53: nit:
```suggestion
        test_types = set.intersection(*(y[item] for item in y))
```
or
```suggestion
        test_types = set.intersection(*(set(y[item]) for item in y))
```

mozci/task.py
```diff
@@ -34,46 +35,37 @@ class Status(Enum):
...
+        self.SUITES = [
+            x
+            for x in test_types
+            if not (
+                x.startswith("talos")
+                or x.startswith("browsertime")
+                or x.startswith("awsy")
+            )
+        ]
+        self.SUITES.sort(reverse=True)
```
marco-c, line 65: Do we need to sort them?

mozci/task.py
```diff
@@ -34,46 +35,37 @@ class Status(Enum):
...
+class Suites:
+    SUITES: set[str] = set()
+
+    def __init__(self):
+        # code used to populate SUITE
+        try:
+            r = requests.get(
+                "https://hg.mozilla.org/mozilla-central/raw-file/tip/taskcluster/ci/test/test-sets.yml"
+            )
+            y = yaml.load(r.text, yaml.BaseLoader)
```
marco-c, line 47: nit:
```suggestion
            test_sets = yaml.load(r.text, yaml.BaseLoader)
```

mozci/task.py
jmaher: yes, so something like `reftest-gpu` is matched before `reftest`

marco-c:
Closing in favor of #1072.

PR #1075, opened 2023-12-05
use task attributes to define configuration instead of parsing the label

jmaher:


mozci/task.py
```diff
@@ -139,6 +141,62 @@ def wpt_workaround(group: str) -> str:
...
+                f"Missing test variant {variant} in variants.yml: {yaml_variants}"
+            )
+        suffixes.append(yaml_variants[variant])
+
+    # figure out which order to pack the suffixes; we can cheat by using self.label
+    retVal = sorted(suffixes, key=lambda x: label.find(x))
+    return "-".join(retVal)
+
+
+def get_configuration(task) -> str:
```
marco-c, line 162: What if, instead of trying to "hack" it into a string, we just use the `task.extra.test-settings` dict as the configuration?

marco-c: Ah I remember we talked about it and we decided to go with treeherder-platform as it would be simpler (to avoid changes in bugbug and in the scheduler in m-c).

marco-c: There are a lot of hacks involved though, more than I was hoping :(

mozci/data/contract.py
```diff
@@ -33,6 +33,8 @@ class Contract:
                     "id": v.Str(),
                     "label": v.Str(),
                     "suite": v.Str(nullable=True),
+                    "platform": v.Str(nullable=True),
+                    "variant": v.Dict(extra=(v.Str(), v.Bool())),
```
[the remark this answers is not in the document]
marco-c: We don't need to modify the treeherder datasource if we do #1077.

mozci/task.py
jmaher: @marco-c is this something we can move forward on?  This gets us out of the label parsing game, but adds more hacks.  We could file bugs to address these issues, which might end up adding hacks to taskcluster code, but at least it would be in one place, and not in more places.

mozci/task.py
```diff
@@ -139,6 +141,62 @@ def wpt_workaround(group: str) -> str:
...
 
 
+def get_test_variant(variants, label) -> str:
+    if not variants:
+        return ""
+
+    suffixes = []
+    yaml_variants = fetch_test_variant_yaml()
+    for variant in variants:
+        if not yaml_variants.get(variant, None):
```
marco-c, line 151: nit: ```suggestion
        if variant not in yaml_variants:
```

mozci/task.py
marco-c: My only fear is that the hacks will not age well... We should find a way to ensure the configurations are correct. Maybe a cron job that gets configurations for m-c pushes and notifies when there's a change in the list?
Any other idea?

marco-c (approved):

mozci/task.py
jmaher: as for the hacks, there are 4 that seems hacky:
1) android platform 7-0 -> 7.0; this is not repeated on other platforms, so a future upgrade has little to no risk
2) android junit doesn't need geckoview added, that is because the test suite is `geckoview-junit`, so the `geckoview` would be redundant.  some options for future proofing:
  a) if `geckoview` not in suite:             retVal += "-geckoview"
  b) change the junit suite name to remove geckoview
3) wpt tags run as tasks;  This seems risky, what if we add another one?  This seems like the right solution is to have a better integration with taskcluster configs and ensure the tag is included as part of the settings.  maybe we call these `wpt-variants` so they are treated similarly to variants, but are not variants in the traditional sense.  Moving forward a bug could be filed for this as a todo and once implemented this could be updated to not be a hack.
4) lastly is xpcshell-msix.  Right now msix is a variant and for mochitest that is true.  For xpcshell there is a need to have [different cli args and paths](https://searchfox.org/mozilla-central/source/testing/mozharness/configs/unittests/win_unittest.py#207).  This hack avoids a label like `xpcshell-msix-msix`;  Some options:
 a) change the suite name; this is super hacky, but less work; maybe `xpcshell-msix` -> `xpcshell-ix`
 b) find a way in mozharness to not have a separate suite with different parameters- do this with a variant and pass in the information from the task definition. 
  c) add a bit in xpcshell harness when using the `--msix` flag it would use different app paths
  
@marco-c  Given this, the questions I have are:
1) what other hacks did I overlook?
2) what would be your desired choice in dealing with these given some options above?

marco-c: That's all really, nothing more.
I think a cron job would be ideal for future-proofing. We could also identify if other hacks become necessary in the future (e.g. tasks are added without the necessary info) and so on. We can file it as a follow-up.

marco-c: And it doesn't have to be automated, we could also set up a team calendar with a montly todo item to check the output is correct :)

PR #1113, opened 2024-05-31
Add a comment to mention when we can remove hack 3 as we have fixed xpcshell + msix to be a proper variant

jmaher:


marco-c:
Oh, I didn't think about a problem: bugbug uses mozci also for "historical" info (~1 year). So we can only remove this hack after 1 year has passed since the fix.

marco-c (approved):
Thanks!

PR #1126, opened 2024-07-09
Migrate CI from Ubuntu 22.04 to Ubuntu 24.04

petemoore:
We intend to remove support for Ubuntu 22.04, and are migrating all worker pools to Ubuntu 24.04. Let us know if that is a problem! Thanks. :-)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -58,7 +58,7 @@ tasks:
       dependencies:
         - { $eval: as_slugid("tests_task") }
       provisionerId: proj-mozci
-      workerType: generic-worker-ubuntu-22-04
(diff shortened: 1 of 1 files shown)


marco-c:
@petemoore it looks like the tasks using this worker type are timing out in pending state.

petemoore:
> @petemoore it looks like the tasks using this worker type are timing out in pending state.

Switching from `c3-standard-8` to `n2-standard-4` in taskcluster/community-tc-config#780 seems to have fixed the issue.

marco-c (approved):

marco-c:
Thanks!

Issue #1145, opened 2025-03-24
add support to identify regressions of builds (build bustage9


Archaeopteryx:
Add support to identify if a build bustage is a regression from a check-in with support to revert the change.

Adding the infrastructure case first because it is shorter:
Apart from a code change issue, a failed build can also be an intermittent issue or an infrastructure issue. The ladder should force a tree closure and repeated retries to build until the issue has been resolved. Suggested intervals: immediately, 5 minutes break, then 15 minutes break between fail and retry.

To identify if a failed build is caused by a check-in:
* Check if the job is either tier 1 or 2: If not (= tier 3), ignore.
* Check if job result is `busted` or `exception`. If not, ignore.
* Check if `extra.treeherder.jobKind` is `build`. If not, ignore. TODO: exclude mobile unit tests.
* Check if this push has already such a build failure (e.g. match for an error message which mentions a source file): Skip the current build failure because the already known one will manage this.
* Check if this push had the same job passing before. If yes, skip regression search. By default, successful builds cannot be rerun unless they are enforced.
* Identify the last successful run of the same job type: What was the last push G which ran the same job type successfully before the push with the failed build B?
* If there are P pushes between G and B:
** If there are less than 100 pushes:
*** If there are less than 20 pushes (default interval of pushes which ran every job type on autoland):
**** Add build task for same job type between G and B if there is not already a build scheduled or running.
**** If requested build job succeeds, narrow down regression range.
**** If requested build fails: check if the error message is similar: build regressions can overlap and a later code change cause a compilation failure earlier in the compilation process, altering the failure message.
***** Start a second bisection for the build failure on the earlier push, narrow regression window for later, initially investigated build failure.
***** Rerun the failed build once to reduce the risk of an intermittent failure.
*** If there are 20 ore more pushes between G and B:
**** Repeat the bisection similar to for the with less than 20 pushes but select the pushes to which to add the build tasks the following way:
***** First iteration: Add build task for every round(sqrt(P)) push.
***** Second iteration: Add build task for every push between last successful build and first failed one.
** If there are more than 100 pushes P between G and B: only check for an intermittent failure by rerunning it, report for manual investigation. The job type likely only runs on demand.

Considerations:
* Avoid bisecting with pgo/shippable builds: These can use small hardware pools for a build step and bisecting would be backlogged by the bottleneck.
* Ignore when past jobs got submitted, started and completed but use the push order to identify their logical order.
[...]

PR #1147, opened 2025-03-31
Store extra.treeherder.jobKind on Task

La0:
Part of #1145

This loads the `extra.treeherder.jobKind` from taskcluster task definition onto the `Task` class.


jmaher (changes requested):
thanks, I prefer these simple changes.

Any chance there is a useful unittest you could add to:
https://github.com/mozilla/mozci/blob/master/tests/test_task.py

La0:
I added a test in `tests/test_data_sources.py` as this is where the data parsing is tested

PR #1148, opened 2025-04-02
WIP: Command to analyze regressions

La0:
Refs #1145

Implements the first few rules, with new support for jobkind #1147

First goal is to implement detection rules and give visual feedback (no automation):

```
$ mozci regression 828a657b93e430f78716c4aa5e6ac9db2ed5372e
+------------------------+---------------------------------------------+-----------+
| ID                     | Label                                       | Result    |
+------------------------+---------------------------------------------+-----------+
| G0_1FUeoSRanYxa1j4CNOw | build-signing-win32/opt                     | exception |
| YCWCtfV4TWW2GkyMLNfjhg | generate-profile-win64-nightlyasrelease/opt | exception |
| TWoTrB4TQ_WYb3Lad4R-cw | build-signing-linux64-aarch64/opt           | exception |
| Ht4KhlCmQ9i9ebunDkoM6g | build-linux64-nightlyasrelease/opt          | exception |
| Gv3SKBZoTuC2sX0SjIRbfw | build-macosx64-nightlyasrelease/opt         | exception |
...
```


mozci/task.py
```diff
@@ -273,6 +273,22 @@ def configuration(self):
...
+
+        # Check if job result is busted or exception
+        if self.result not in ("busted", "exception"):
+            return False
+
+        # Check if extra.treeherder.jobKind is build
+        if self.job_kind != "build":
+            return False
+
+        return True
```
marco-c, line 290: We should add this logic within https://github.com/mozilla/mozci/blob/4c9faf8fa5ccdb2933125182d5d89b2dde2d9f34/mozci/push.py#L1128 (which is implementing classification for tests).
We could add another function before that called `classify_builds`, and call it within `classify_regressions`.

This way we can re-use the later logic for retriggers and backfills (which we might need to update).

mozci/console/commands/regression.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+
+from cleo.commands.command import Command
+from cleo.helpers import argument
+
+from mozci.push import Push
+
+
+class RegressionCommand(Command):
```
marco-c, line 9: We could re-use ClassifyCommand, maybe add an option to limit to builds only so it's faster to iterate, and add an option to just visualize results

vrigal:
I added treeherder API helpers to `mozci.util.treeherder` so it is easier to iterate/review at first. I noticed I couldn't fetch a job for every task in Bastien's example, maybe it is expected ?
```
14:35:18.499 | INFO     | mozci.util.treeherder:74  - Linked 50/393 tasks to jobs
14:35:18.499 | WARNING  | mozci.util.treeherder:78  - Some tasks could be linked to Treeherder jobs
```

La0:
Closing in favor of #1151

PR #1151, opened 2025-04-14
List similar tasks from previous pushes to identify regressions

vrigal:
References #1145
Based on #1148

PR #1157, opened 2025-05-16
Identify build regression in the classify command

vrigal:
Based on #1151

Preserves the command to lisk tasks to retrigger (dry run).
Retriggers tasks matching build failure filters in the `push classify` command.


vrigal:
@marco-c I took the work from #1151 and added 1 commit on top of it.
My initial idea was to keep the generic analysis and retrigger the detected build tasks. Also I added a flag `--disable-build-regressions` to be able to switch to the current behavior from the config.

mozci/console/commands/push.py
```diff
@@ -264,7 +295,7 @@ class ClassifyCommand(Command):
         ),
         option(
             "environment",
-            description="Environment in which the analysis is running (testing, production, ...).",
+            description="Environment in which the analysis is running (testing, production, ...). ",
```
marco-c, line 298: nit: remove whitespace

mozci/console/commands/push.py
```diff
@@ -314,7 +342,25 @@ def handle(self) -> None:
...
+                    else:
+                        self.line(
+                            f"Retriggering {len(tasks_to_retrigger)} build regression that may introduce a build bustage"
+                        )
+                        for task in tasks_to_retrigger:
+                            task.retrigger(push)
+                except Exception as e:
+                    self.line(
+                        f"<error> Couldn't run build regressions analysis on push {push.push_uuid}: {e}.</error>"
+                    )
```
marco-c, line 362: As a follow-up, we could later just merge this into `classify_regressions` and add to the `to_retrigger_or_backfill` list

mozci/console/commands/build_regressions.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: Let's move all the functions from this file into push.py, before https://github.com/mozilla/mozci/blob/917898bfc6b55ee6d4a3b207ef4d6c014e3c39b8/mozci/push.py#L1128

mozci/console/commands/push.py
vrigal: I opened PR #1158.
We should be able to rebase once we are sure it works with the try-catch behavior and the option to disable it.

marco-c (approved):

PR #1163, opened 2025-05-22
Rules for retriggering build tasks with previous occurrences

vrigal:


marco-c:
@Archaeopteryx @jmaher need your input here.
When do we want to retrigger a build that has previous failure occurrences and when don't we?

marco-c:
> fixed by commit - retrigger would most likely fail again

I guess the only case where retriggering would be helpful is:
Push 1 - Build failed, fixed by commit in Push 2
Push 2 - Build doesn't run
Push 3 - Build fails

This way we can confirm it is actually fixed. I'm not sure how frequent this is.

Another interesting case would be:
Push 1 - Build failed, fixed by commit in Push 4
Push 2 - Build doesn't run
Push 3 - Build fails
Push 4

In this case, Push 3 might not need to retrigger the build because we already know it's fixed in a later push.

Archaeopteryx:
This sounds reasonable.

In general,
* a failure needs to be checked if it's intermittent or permanent - the easiest is a retrigger which shall produce the same error message;
* the start of a regression shall be found, with retriggers and backfills if necessary
* the end of a regression shall be verified by running the task for newer pushes after the latest failure - at some point the task shall succeed for newer pushes, it could have a different failure message from another issue before being successful

vrigal:
@marco-c I did a first implementation of the mentioned rules.
I think adding logs and writing tests can help much on this part. I'll have a look at it next week.

mozci/task.py
```diff
@@ -423,14 +423,78 @@ def is_build_failure(self) -> bool:
...
+            (task for task in previous_tasks if task.failed), None
+        )
+
+        # 2. No previous occurrence have a classification
+        if last_failure_occurrence is None:
+            return True
+
+        # 3. Last failure occurrence is classified as intermittent
+        if last_failure_occurrence.is_intermittent:
+            return True
```
marco-c, line 483: This one too, because if summary.is_intermittent is True, `prior_regression` stays False.

mozci/task.py
```diff
@@ -423,14 +423,78 @@ def is_build_failure(self) -> bool:
...
+        ):
+            return True
+
+        last_failure_occurrence = next(
+            (task for task in previous_tasks if task.failed), None
+        )
+
+        # 2. No previous occurrence have a classification
+        if last_failure_occurrence is None:
+            return True
```
marco-c, line 479: This actually is "there is no previous occurrence", so it is redundant with "count == 0", you can remove it. The 4th rule is actually implementing `No previous occurrence have a classification`

mozci/task.py
```diff
@@ -423,14 +423,78 @@ def is_build_failure(self) -> bool:
...
+        if last_failure_occurrence.classification == "not classified":
+            return True
+
+        # 5. There is only one task on the current push with the same name and failed status
+        if (
+            len(push.tasks) == 1
+            and last_failure_occurrence.label == self.label
+            and last_failure_occurrence.result == self.result
+        ):
+            return True
```
marco-c, line 495: This rule actually applies on top of the others. We should make sure to only retrigger a build once.
You could move it at the top, right after the `tier` check, and just return False when you find another task in the same push with the same name.

something like:
```Python
if sum(1 for task in push.tasks if task.label == self.label) > 1:
    return False
```
(you don't need the `result` part because this function is always only called on tasks that failed)

Or you could implement this in the check_build_regressions function, filtering `tasks_to_retrigger`:
```Python
tasks_to_retrigger = [task for task in tasks_to_retrigger if sum(t.label == task.label for t in push.tasks) == 1]:
```

mozci/task.py
```diff
@@ -423,14 +423,78 @@ def is_build_failure(self) -> bool:
...
+        if last_failure_occurrence is None:
+            return True
+
+        # 3. Last failure occurrence is classified as intermittent
+        if last_failure_occurrence.is_intermittent:
+            return True
+
+        # 4. Last failure occurrence is not classified yet
+        if last_failure_occurrence.classification == "not classified":
+            return True
```
marco-c, line 487: This case is not yet covered by `get_regressions`. Since it is the only one though, maybe we could just change `get_regressions` and in this function return True independently from the value of `previous_occurrences_count`.

marco-c: Basically we'd just need to change:
```Python
                            if (
                                summary.status != Status.PASS
                                and not summary.is_intermittent
                            ):
                                prior_regression = True
```
to be something like:
```Python
                            if (
                                summary.status != Status.PASS
                                and not summary.is_intermittent
                                and summary.classification != "not classified"
                            ):
                                prior_regression = True
```

Perhaps we can add a parameter to `get_regressions` to enable this behavior so current users won't get affected by the change. The parameter could be named `assume_unclassified_is_intermittent`

Archaeopteryx (approved):

mozci/push.py
```diff
@@ -982,7 +982,10 @@ def find(runnable_type: str) -> Optional[str]:
 
     @memoize
     def get_regressions(
-        self, runnable_type: str, historical_analysis: bool = True
+        self,
+        runnable_type: str,
+        historical_analysis: bool = True,
+        assume_unclassified_is_intermittent: bool = False,
```
marco-c, line 988: You could add a test for this parameter in https://github.com/mozilla/mozci/blob/master/tests/test_regressions.py.

A quick way to do it would be to:
- temporarily set this to True by default;
- run tests/test_regressions.py;
- get the failures;
- add the failing tests as new tests with adjusted expectations.

mozci/push.py
```diff
@@ -993,6 +996,8 @@ def get_regressions(
         last time the task passed (so any one of them could have caused it).
         A count of MAX_DEPTH means that the maximum number of parents were
         searched without finding the task and we gave up.
+        If the parameter `assume_unclassified_is_intermittent` is set to True, previous
+        failures that have not been classified will be considered as regressions.
```
marco-c, line 1000:
```suggestion
        If the parameter `assume_unclassified_is_intermittent` is set to True, previous
        failures that have not been classified will be considered as intermittent.
```

mozci/push.py
```diff
@@ -1042,6 +1047,13 @@ def get_regressions(
                             if (
                                 summary.status != Status.PASS
                                 and not summary.is_intermittent
+                                and (
+                                    not assume_unclassified_is_intermittent
+                                    or all(
+                                        c != "not classified"
+                                        for c, n in summary.classifications
+                                    )
```
vrigal, line 1055: Thank you much for the reviews.
I had to change this condition because of the way tasks are stored as "runnable_summaries". I don't fully understand this part, but it make sense according to the module and tests.

marco-c: Basically, "runnable" is an abstraction. We have both "groups of tests" and "tasks", and "runnable" is the abstraction for them.

marco-c (approved):

PR #1167, opened 2025-06-03
Avoid retrigger a task failed with exception DEADLINE_EXCEEDED

### LINKED ISSUE #1165: Don't retrigger builds which failed with exception because of DEADLINE_EXCEEDED
If they fail this way, it means most likely there was a dependency that didn't run.

vrigal:
Closes #1165

### DIFF
--- a/mozci/data/contract.py
+++ b/mozci/data/contract.py
@@ -54,6 +54,7 @@ _contracts: Tuple[Contract, ...] = (
                             "exception",
                             "canceled",
                             "superseded",
+                            "deadline-exceeded",
                         ]
                     ),
                     "job_kind": v.Str(),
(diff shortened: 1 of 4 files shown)


vrigal:
I tested with revision `88781efc5a6e91ec9b485b7aa01505a8879da58c` (from https://github.com/mozilla/mozci/issues/1165#issuecomment-2935103479):
* before:                   
```
Fetched 308 tasks for push 231918.
Detected 27 build tasks that may contain a regression (22 were never seen before).
22 tasks have been detected as potential regressions:
  […]                     
  * repackage-signing-msix-win64/debug
```
* after:
```
Fetched 308 tasks for push 231918.
No build regression detected.
```

PR #1185, opened 2025-09-19
update path for variants.yml about test variants

Archaeopteryx:
mozilla-central change in bug 1982404


marco-c (approved):

marco-c:
Ah, we also need to update the variants.yml URL in tests/test_push.py.

PR #1190, opened 2025-09-25
Prevent retriggering more than 15 failed build tasks

### LINKED ISSUE #1188: Retrigger failed tier-1 build tasks once
Definition of "failed build tasks":
* Result "busted" shown in Treeherder. This is the same as `jobKind` of the `extra` payload of the task in Taskcluster with a build [not equal](https://github.com/mozilla/treeherder/blob/81a0387e3a8b829a087e232d0f55f1e7541dde62/treeherder/etl/job_loader.py#L227) to `test`.
* Information, covered by previous point: "test" not mentioned in task name. Some tasks of the build process do not contain "build", e.g. [generate-profile-win64-aarch64-shippable/opt](https://firefox-ci-tc.services.mozilla.com/tasks/GOf6Ig8nRbC11Q4Qf7CSRQ)

Limit total reruns per push
In case many build failures are observed, the issue is likely not general intermittentness of the tasks if the tree/repository/project is mozilla-central but either a permanent failure of the executed task payload or an infrastructure, e.g. to access required resources. Until a detection for the latter has been implemented, only up to 15 failed build tasks in a single push shall be rerun. This should also be done when it is known there are more than 15 such task failures to enable easier human investigation later.
It is recommended to design this in a way which allows the evaluation if a task shall be rerun by tree/repository name and an input variable of the maximum reruns per push.

@vrigal @La0

vrigal:
Closes #1188

The two initial points about build failure detection seems to be covered by #1167. The detection is based on the `jobKind` attribute being `"build"`, and result being one of `{"busted", "failed", "exception"}` .

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -317,7 +317,10 @@ class ClassifyCommand(BasePushCommand):
         ),
         option(
             "maximum_build_tasks_to_retrigger",
-            description="The maximum number of build task failures to retrigger for a single push",
+            description=(
+                "The maximum number of build task failures to retrigger for a single push. "
+                "The value can be set to -1 for no maximum."
+            ),
             flag=False,
             default=15,
         ),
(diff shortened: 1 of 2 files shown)


vrigal:
> It is recommended to design this in a way which allows the evaluation if a task shall be rerun by tree/repository name and an input variable of the maximum reruns per push.

@Archaeopteryx could the tree/repository filter be implemented at the "infrastructure" level (when calling the push classification command) ? The parameter can be set to a value of `-1` (no maximum of retriggered task).

vrigal:
I force-pushed because a CI error occurred while installing the virtualenv:
https://community-tc.services.mozilla.com/tasks/VGSQuHfWQu2X0lVE1rkW3w/runs/0/logs/public/logs/live.log#L439

mozci/console/commands/push.py
```diff
@@ -350,6 +359,18 @@ def handle(self) -> None:
                     if not tasks_to_retrigger:
                         self.line("No build task detected as potential regression")
                     else:
+                        max_retrigger = self.option("maximum_build_tasks_to_retrigger")
+                        if max_retrigger > 0:
```
Archaeopteryx, line 363: `0` would be a valid value to disable the feature. Should the check be for `> -1`?

Archaeopteryx: On the other hand, if were no retriggers supposed to get scheduled, a warning is not needed if the user is aware.

Archaeopteryx:
> > It is recommended to design this in a way which allows the evaluation if a task shall be rerun by tree/repository name and an input variable of the maximum reruns per push.
> 
> @Archaeopteryx could the tree/repository filter be implemented at the "infrastructure" level (when calling the push classification command) ? The parameter can be set to a value of `-1` (no maximum of retriggered task).

Yes.

Archaeopteryx (approved):

PR #1193, opened 2025-09-26
Use the right actions to retrigger tasks

### LINKED ISSUE #1192: mozci used the `retrigger-decision` action for a non-decision task
Example: https://firefox-ci-tc.services.mozilla.com/tasks/XlTQ5VQDR6OiB6CEzCIPRA

Quite clearly shows it was created by MozCI: `Action triggered by clientID project/mozci/sheriff-automation`

And we had: `  ACTION_CALLBACK: retrigger-decision`

Image

This ended up doing something totally unsupported, and confusing.

marco-c:
Fixes #1192

### DIFF
--- a/mozci/task.py
+++ b/mozci/task.py
@@ -312,13 +312,27 @@ class Task:
         sig = signature(self.__init__)
         return {k: v for k, v in self.__dict__.items() if k in sig.parameters}
 
+    def _ensure_context(self, action):
+        # An empty context means the action is relevant to the task-group.
+        if len(action["context"]) == 0:
+            return True
+
+        for potential_matcher in action["context"]:
+            if all(
+                self.tags.get(key) == value for key, value in potential_matcher.items()
+            ):
+                return True
+
+        return False
+
     def _get_action(self, decision_task, action_name):
         actions = decision_task.get_artifact("public/actions.json")
-        action = next(
-            action for action in actions["actions"] if action["name"] == action_name
-        )
-        assert action["kind"] == "hook"
-        return action
+        for action in actions["actions"]:
+            if action["name"] == action_name and self._ensure_context(action):
+                assert action["kind"] == "hook"
+                return action
+
+        return None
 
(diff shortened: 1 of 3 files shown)


jcristau (dismissed):
Per my comment on the issue I don't think this is correct.

mozci/task.py
```diff
@@ -315,7 +315,9 @@ def to_json(self):
     def _get_action(self, decision_task, action_name):
         actions = decision_task.get_artifact("public/actions.json")
         action = next(
-            action for action in actions["actions"] if action["name"] == action_name
+            action
+            for action in actions["actions"]
+            if action["hookPayload"]["decision"]["action"]["cb_name"] == action_name
```
jcristau, line 320: I don't think matching cb_name is right.

mozci/task.py
```diff
@@ -339,47 +341,57 @@ def _trigger_action(self, action, payload):
...
-        force_retrigger parameter is set to True.
-        """
-        if not force_retrigger and self._should_retrigger() == "false":
-            logger.info(
-                "Not retriggering task '{}', task should not be retriggered".format(
-                    self.tags.get("label")
-                )
-            )
-            return None
+    def _ensure_context(self, action):
```
jcristau, line 344: This should be done as part of _get_action probably, since it's relevant to which action to select.

mozci/task.py
marco-c: How should we match?
Are there cases where multiple actions with the same name could have context that applies to a given task?

jcristau: Yes: "The order of the array of actions is significant: actions should be displayed in this order, and when multiple actions apply, the first takes precedence", and "The `name` is used by user interfaces to identify the action".

mozci/task.py
marco-c: Done.

mozci/task.py
marco-c: All right, fixed.

Issue #1191, opened 2025-09-25
retrigger failed test tasks for Linux cloud pools 5 times and identify permanent failures


Archaeopteryx:
Subtask of #1189 

If a task failed and has the `jobKind` of `test`:
* Check if the task has been classified. If yes, do nothing and quit here.
* Check if the task is tier 1 or 2. If not, do nothing and quit here.
* Identify if the failure are test failures or other issues, e.g. infrastructure
  * If there is no task artifact which contains `errorsummary` in its name or is incomplete, then the task either hit an infrastructure issue or the test suite does not support failure summaries by `errorsummary`.
* If test failure
  * Check if the task got executed on Linux cloud worker pool. The Task Queue ID listed for it ([example](https://firefox-ci-tc.services.mozilla.com/tasks/Lq7BnsF2TM-FhhlgLEyadw) should start with `gecko-t/t-linux-`.
  * Schedule up to `5` task retriggers to have at least 6 tasks (original one + 5 retriggers): check if there are already retriggers scheduled for the push
* retriggers complete: Decide if the failure rate is a regression
  * retriggers: At least 2 retriggers observed the same test failures as the initial task: regression

Classification would be automatically by LPID.

vrigal:
@Archaeopteryx I make a first draft implementation in #1196, so we can discuss the details there.

vrigal:
@Archaeopteryx do you have an idea how to compare if a failed retrigger is the same issue than the original task ?
We don't have access to the logs from Mozci, and I suppose it's not a reliable way to compare the failure.

Should we consider any retrigger failure as a similar failure ?

Archaeopteryx:
Recommendation:

1. For the initially failed task, fetch the `bug_suggestions` data from Treeherder ([example](https://treeherder.mozilla.org/api/project/mozilla-central/jobs/530611527/bug_suggestions/)).
2. Get all array elements (= failure lines) for which `search` starts with `TEST-UNEXPECTED-`.
3. Count how often these failures have been observed in the retriggers. This likely needs some delay because Treeherder needs to parse the logs first. The alternative is to use the .*errorsummary.*\.json artifact and only compare the failed manifests (test folders) but it has the risk to be too noisy, especially for platforms where tests execute slowly (ASan, TSan).

PR #1196, opened 2025-09-26
Retrigger failed test tasks to detect permanent ones

vrigal:
Part 1/2 of #1191

Adds up to 5 retriggers to failed tasks on queue `gecko-t/t-linux-`, for each test tasks with failure (tier 1 or 2, not classified).

### DIFF
--- a/mozci/console/commands/push.py
+++ b/mozci/console/commands/push.py
@@ -408,7 +408,7 @@ class ClassifyCommand(BasePushCommand):
                 continue
 
             # Rettriger test failures to detect permanent ones, only on Linux cloud worker pool for now
-            push.identify__permanent_failures(queue_prefix="gecko-t/t-linux-")
+            push.identify_permanent_failures(queue_prefix="gecko-t/t-linux-")
 
             if self.option("show-intermittents"):
                 self.line("-" * 50)
--- a/mozci/console/commands/batch_execution.py
+++ b/mozci/console/commands/batch_execution.py
@@ -93,7 +93,7 @@ def run_combinations_for_push(push):
         try:
             classification, regressions = push.classify(**parameters)
             # Rettriger test failures to detect permanent ones, only on Linux cloud worker pool for now
-            push.identify__permanent_failures(queue_prefix="gecko-t/t-linux-")
+            push.identify_permanent_failures(queue_prefix="gecko-t/t-linux-")
             end = time.time()
 
             # Only save results to a JSON file if the execution was successful


mozci/console/commands/push.py
```diff
@@ -407,6 +407,9 @@ def handle(self) -> None:
                 self.line(f"<error>{traceback.format_exc()}</error>")
                 continue
 
+            # Rettriger test failures to detect permanent ones, only on Linux cloud worker pool for now
```
vrigal, line 410: Do you think it is ok to run the identification at the push level, right after calling `classify` ?

mozci/push.py
```diff
@@ -1440,6 +1440,33 @@ def _map_failing_tasks(groups):
...
+                and task.queue_id.startswith(queue_prefix)
+                and task.can_retrigger_test_failure()
+            )
+        ]
+        logger.info(
+            f"Detected {len(failures_to_retrigger)} test failures that can be retriggered to identify permanent failures"
+        )
+        for failure in failures_to_retrigger:
+            """
+            Steps to be implemented:
```
vrigal, line 1462: @Archaeopteryx  I would like to have your input on the following steps.

mozci/console/commands/push.py
Archaeopteryx: Yes, good to run it at the push level.

mozci/push.py
vrigal: > This needs a query for jobs of the same push and with the same job_type__name (or job_type__id), e.g. with the Treeherder API and preferably cached.

Don't we already have fetched thos retriggered tasks at this point ? The comment of `Push.tasks()` explicitly states "All tasks that ran on the push, including retriggers and backfills": https://github.com/mozilla/mozci/blob/master/mozci/push.py#L327.

The second part seems more complex to implement. Should I first focus on retriggering the tasks 5 times, and for example always consider regression if all of the retriggers failed ? I think Marco suggested doing this development in two PRs.

Also I have one more question: How can I "backfill a task" from mozci ? Should I use the Treeherder API to publish something in particular ?

Archaeopteryx: > Don't we already have fetched those retriggered tasks at this point ? The comment of Push.tasks() explicitly states "All tasks that ran on the push, including retriggers and backfills": https://github.com/mozilla/mozci/blob/master/mozci/push.py#L327.

In this case ignore the fetching and only get the tasks of similar type from the mozci cache.

> Do you think the retriggers should be started one by one at each push analysis ?

If a task fails, the five retriggers shall be requested to run in parallel

> The second part seems more complex to implement. Should I first focus on retriggering the tasks 5 times, and for example always consider regression if all of the retriggers failed ?

For the initial implementation here without backfills, just doing the retriggers is good. When the backfills are added in later PR, we want to avoid backfills for permanent failures which have not been added by a recent push: They could be from changes to the CI (machines, toolchains), tasks which run as tier 2 and don't get backed out for that but wait for a fix.

> Also I have one more question: How can I "backfill a task" from mozci ? Should I use the Treeherder API to publish something in particular ?

Backfils should be done with the [current interface](https://github.com/mozilla/mozci/blob/e192e0145e7f4ccaa750c6dfb70fa4ec87a43724/mozci/task.py#L406). Nothing is needed on Treeherder's end, it will ingest the tasks automatically like tasks scheduled by default.

vrigal:
@Archaeopteryx I'm starting the work on the second part, implementing the backfills for the cases we discussed, on top of this PR.

vrigal:
> @vrigal we can implement that in a separate PR, so we can review and land this one faster

In a separate PR for sure! Sorry it wasn't very explicit.

Archaeopteryx (approved):

PR #1208, opened 2025-11-04
Backfill failed retriggered tests based on Treeherder bug suggestions

### LINKED ISSUE #1202: backfill permanent or new highly frequent test failures
If a test task has been identified as
a) either a permanent failure with the same issue
b) or a frequent test failure (>=50% of runs - that's the original one +5 retriggers) which had not been observed before: For the initial version, only treat it as new if there is no intermittent bug for it.
Backfill the task: Add it to previous pushes where the test manifest had not been executed.
For permafails: Run it once.
For frequent failures: Run it 6 times which is similar to the push for which it had been identified initially.

Out of scope: Early identification of regressing push when last good and first bad revision could be identified but backfilled tasks are still running for some pushes.

vrigal:
Part 2/2 of https://github.com/mozilla/mozci/issues/1191

Closes #1202

### DIFF
--- a/mozci/push.py
+++ b/mozci/push.py
@@ -1550,7 +1550,14 @@ class Push:
                 for suggestions in bug_suggestions
                 if suggestions["search"].startswith("TEST-UNEXPECTED-")
             }
-            if not terms:
+            if not terms and task == failure:
+                # The initial failure should return at least 1 valid error line to continue with the analysis
+                logger.warning(
+                    "No failure line in standard format with TEST-UNEXPECTED- found for initial task "
+                    f"{failure.label} ({failure.id}) in the corresponding Treeherder job {job['id']}, skipping."
+                )
+                return
+            elif not terms:
                 logger.debug(
                     f"No failure line in standard format with TEST-UNEXPECTED- found for job {job['id']}"
                 )
@@ -1570,12 +1577,6 @@ class Push:
 
         # Analyze the most common failure lines across the initial task and the retriggers
         most_common_failure = Counter(overall_failure_lines).most_common(1)
-        if not most_common_failure:
-            logger.info(
-                f"Test failure {failure.label}: "
-                f"No retrigger match the initial failure, nothing to do."
-            )
-            return
 
         value, counter = most_common_failure[0]
         logger.info(


mozci/push.py
```diff
@@ -1487,12 +1489,79 @@ def identify_permanent_failures(self, retriggers_count=5, queue_prefix="") -> No
...
+            return
+
+        if matching_retriggers >= 2:
+            # >=50% of runs (initial + 2 out of 5 retriggers) is a frequent test failure and should be backfilled
+            logger.info(
+                f"Test failure {failure.label}: "
+                "Backfilling initial task as >= 50% of the runs caused the same failure "
+                f"({matching_retriggers + 1} out of {len(retriggers) + 1})."
+            )
+            failure.backfill(self)
```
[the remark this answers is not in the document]
Archaeopteryx: Sorry. The backfilled tasks are similar to retriggers, there should be no issue with multiple backfills (except the risk of overscheduling if the two action tasks are not taken into account).

mozci/push.py
```diff
@@ -1440,10 +1441,30 @@ def _map_failing_tasks(groups):
...
+        if len(values) != 3:
+            logger.warning(
+                f"Invalid failure term '{term}': considering the failure as new."
+            )
+            return False
+        _, full_path, info = values
+        # Consider that an intermittent bug exist for this issue based on its name + status
+        return any(
+            bug["status"] in ("NEW", "REOPENED")
+            and bug["summary"] == f"Intermittent {full_path} | single tracking bug"
```
Archaeopteryx: Please do an `ends_with` check instead without `Intermittent`. Sometimes these bugs get prefixed to indicate restrictions (e.g. `Intermittent Linux debug`, `Beta Intermittent`.

mozci/data/sources/treeherder/__init__.py
```diff
@@ -97,3 +102,32 @@ def run_pushes(self, branch, nb=15) -> List[Dict]:
...
+    def check_logs_parsed(self, job_id: int, branch: str = "autoland") -> bool:
+        """Check if the logs of a specific job have been parsed already.
+        Raises JobUnavailable in case no job log are found or some are still in a pending state.
+        """
+        data = self._run_query(f"/project/{branch}/job-log-url/?job_id={job_id}")
+        if not data:
+            raise JobUnavailable
+        if any(job_log_url.get("parse_status") == "pending" for job_log_url in data):
+            raise JobUnavailable
+        return all(job_log_url.get("parse_status") == "parsed" for job_log_url in data)
```
Archaeopteryx: The return value is not used - is this intended?

mozci/push.py
```diff
@@ -1487,12 +1508,88 @@ def identify_permanent_failures(self, retriggers_count=5, queue_prefix="") -> No
                 )
 
             to_retrigger = retriggers_count - len(retriggers) - scheduled_retriggers
-            if to_retrigger <= 0:
-                # TODO: Analyze the result of the retriggers to classify between backfills and permanent failures
+            if to_retrigger <= 1:
```
Archaeopteryx: Why the change from `0` to `1`? This way the retrigger won't get scheduled (line 1516).

mozci/push.py
```diff
@@ -1487,12 +1508,88 @@ def identify_permanent_failures(self, retriggers_count=5, queue_prefix="") -> No
...
+        if not terms:
+            logger.warning(
+                f"No bug suggestion related to the failure {failure.label} could be fetched from Treeherder, skipping."
+            )
+            return
+        # Filter out failures that already have an intermittent bug opened
+        terms = {
+            term: open_recent
+            for term, open_recent in terms.items()
+            if not self.has_intermittent_bug(term, open_recent)
```
Archaeopteryx: Initially, this could be done for new failures (`not ... has_intermittent_bug`). But it is likely better to ignore if bugs or internal issues have been been filed in the past and instead:

- If it's a permanent failure: backfill with one task per push.
- If it's intermittent with >50% of tasks on the same push with a shared failure message: backfill with 6 tasks per push.

mozci/push.py
```diff
@@ -1487,12 +1508,88 @@ def identify_permanent_failures(self, retriggers_count=5, queue_prefix="") -> No
...
+        matching_retriggers = sum(
+            1
+            for retrigger in retriggers
+            if (
+                retrigger.state == failure.state
+                and all(
+                    term
+                    in retrigger.get_artifact("public/logs/live.log").content.decode()
+                    for term in terms.keys()
+                )
```
Archaeopteryx: This assumes every failure line of the initially failed tasked is part of the regression which might not be true. A `Counter` of failure lines and a check for counts >= 3 would also justify backfills.

mozci/data/sources/treeherder/__init__.py
vrigal: I made the changes but still have a question. As this check only concern the main failure, could such states affect the `bug_suggestion` response from Treeherder ?

mozci/push.py
```diff
@@ -1487,12 +1508,88 @@ def identify_permanent_failures(self, retriggers_count=5, queue_prefix="") -> No
...
+            return
+
+        if matching_retriggers >= 2:
+            # >=50% of runs (initial + 2 out of 5 retriggers) is a frequent test failure and should be backfilled
+            logger.info(
+                f"Test failure {failure.label}: "
+                "Backfilling initial task as >= 50% of the runs caused the same failure "
+                f"({matching_retriggers + 1} out of {len(retriggers) + 1})."
+            )
+            failure.backfill(self)
```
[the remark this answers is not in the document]
vrigal: I suppose "permanent" here means that all of the retriggered tasks matched the failure ?

mozci/data/sources/treeherder/__init__.py
Archaeopteryx: There cannot be bug suggestions for `failed` and `skipped_size`, behavior should be unchanged.

mozci/push.py
```diff
@@ -1483,16 +1504,111 @@ def identify_permanent_failures(self, retriggers_count=5, queue_prefix="") -> No
...
+                if not self.has_intermittent_bug(term, open_recent)
+            }
+            if not terms:
+                logger.debug("All the failures already have a bug opened, skipping.")
+                continue
+            overall_failure_lines.extend(set(terms))
+
+        # Analyze the most common failure lines across the initial task and the retriggers
+        most_common_failure = Counter(overall_failure_lines).most_common(1)
+        if not most_common_failure:
```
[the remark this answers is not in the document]
vrigal: I think it will work as you explained in most cases.
I removed this check, and added a condition before to ensure the initial task returned at least 1 valid failure line.

